PR #131, opened 2023-03-24
Add version attribute to providers

wagoodman:
This sets a `__version__` attribute in the base class that can be optionally overridden for provider implementations that need to convey a breaking change to the semantics of the data being written.


DEVELOPING.md
```diff
@@ -285,6 +285,11 @@ Possible vulnerability schemas supported within the vunnel repo are:
 - [GitHub Security Advisories](https://github.com/anchore/vunnel/tree/main/schema/vulnerability/github-security-advisory)
 - [NVD Vulnerability](https://github.com/anchore/vunnel/tree/main/schema/vulnerability/nvd)
 
+If at any point a breaking change needs to be made to a provider (and say the schema remains the same), then you
+can set the `__version__` attribute on the provider class to a new integer value (incrementing from `1` onwards). This
```
spiffcs, line 289: Do you just want these to be 1, 2, 3, 4?

If there is no breaking change does it matter to keep track of additions to the structure?

spiffcs (approved):
Looks good! Just a question about how you want the versions to be represented since some of the fields have semver while the actual value is 1 and only incremented on a breaking change

wagoodman:
The only field where the version is being set for the provider is the __version__ attribute, which is an integer (the breaking field in a semver). The data shape output by a provider is controlled by a schema semver. The fun part about this is that bumping the schema in a breaking way doesn't necessarily mean that old and new results can't be mixed (since each individual record has the schema url referenced). It's up to the provider to decide if the change is breaking. Downstream in grypedb as long as there is an unmarshaller for that schema version it will handle it appropriately.

PR #132, opened 2023-03-27
Add provider for Chainguard Linux

luhring:
Adds a provider for Chainguard Linux. Additionally has misc. fixes in the documentation.

### DIFF
--- a/tests/quality/config.yaml
+++ b/tests/quality/config.yaml
@@ -18,7 +18,7 @@ yardstick:
       # Note:
       #  - ALWAYS leave the "import-db" annotation as-is
       #  - this version should ALWAYS match that of the other "grype" tool below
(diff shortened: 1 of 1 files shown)


src/vunnel/providers/chainguard/__init__.py
```diff
@@ -0,0 +1,64 @@
...
+            result_store=result.StoreStrategy.SQLITE,
+            existing_results=provider.ResultStatePolicy.DELETE_BEFORE_WRITE,
+        ),
+    )
+    request_timeout: int = 125
+
+
+class Provider(provider.Provider):
+    _url_ = "https://packages.cgr.dev"
+    _db_types = ["chainguard"]
```
wagoodman, line 28: I think this should be deleted (it's really in Parser it seems)

wagoodman (approved):

PR #135, opened 2023-03-28
Fix grype-db install within quality gate

wagoodman:
Currently it isn't possible to install grype-db from a fork. This PR refactors the grype-db install code to be a little more modular and adds install capability for forks and local user paths (e.g. `file://~/code/grype-db`).


kzantow (approved):
👍

PR #167, opened 2023-05-02
feat: use RHEL v2 OVAL files where possible

westonsteimel:
Updates the RHEL provider to use the v2 OVAL files when available


src/vunnel/providers/centos/parser.py
```diff
@@ -58,96 +58,166 @@
...
-    _xml_sha_file_ = "com.redhat.rhsa-all.xml.sha256sum"
+    _url_mappings_ = [
+        # Legacy data for RHEL:5
+        {
+            "base_url": "https://www.redhat.com/security/data/oval",
+            "manifest_path": "PULP_MANIFEST",
+            "oval_paths": [
+                "com.redhat.rhsa-all.xml.bz2"
+            ],
+        },
```
westonsteimel, line 69: I'm not sure yet what to do here.   There doesn't appear to be data for RHEL 5 under the v2 api, so for now we still need to get it from the old dataset.  That will be turned off in June, but the archive location it will go to isn't up yet so we can't yet switch over to pulling from there

westonsteimel: Maybe we just rely on that file being in the prebuild and don't attempt to download it anymore?

westonsteimel: Or attempt to download but if it fails just use the latest local version for it?

wagoodman: I think that's the best thing to do for now -- this will be archived data in the OSS cache + Enterprise preload.

wagoodman (approved):

PR #171, opened 2023-05-03
chore: more info logging for rhel long operations

westonsteimel:
When a full sync of the RHEL CVE data is triggered, there is no log output for a very long time at info level, so this at least outputs a row for each page of data processed, which should indicate that things are progressing


wagoodman (approved):
>             self.logger.info("full sync triggered, downloading all CVEs.  This may take quite some time.")

#logthetruth

PR #181, opened 2023-05-17
Add mariner

willmurphyscode:
Implements one approach to adding Mariner vulnerability feed, see https://github.com/anchore/grype/issues/1220.

Wanted to get some feedback on a couple of things:

1. This uses `xsdata`, a new dependency, to generate dataclasses representing the OVAL XML found in the repo
2. This uses `lxml`'s element tree and XPath's to pull apart the XML before handing it off to `xsdata` for parsing
3. 1 and 2 are a new pattern, implemented independently of the OVAL definition work
4. 1 and 2 have a bit of a hack between them, namely that, in order to get item by item handling out of xsdata, we repeatedly parse partial documents, so there's a bit of redundant serialization/deserialization going on.

There are also some things that are probably missing:
1. The update strategy - in this draft PR, the downloader just always runs. We need to figure out a better strategy there.
2. The XML we observe over at https://github.com/microsoft/CBL-MarinerVulnerabilityData is narrower than the OVAL XML specification allows. (For example, every `criteria` node seems to have `AND` as its operator.) We might need to talk about future proofing. The current revisions future proofing is basically done by parsing the XML one CVE at a time and logging failures.
3. The config object for the new provider is basically copy/pasted; more configuration might be needed than this implements.


src/vunnel/providers/mariner/__init__.py
```diff
@@ -0,0 +1,77 @@
...
+    # api_key: str = "env:NVD_API_KEY"
+
+    def __post_init__(self) -> None:
+        pass
+        # if self.api_key.startswith("env:"):
+        #     self.api_key = os.environ.get(self.api_key[4:], "")
+
+    def __str__(self) -> str:
+        # sanitize secrets from any output
+        # api_value = self.api_key
```
wagoodman, line 33: dead code -- should probably be deleted

src/vunnel/providers/mariner/__init__.py
```diff
@@ -0,0 +1,77 @@
...
+        default_factory=lambda: provider.RuntimeConfig(
+            result_store=result.StoreStrategy.SQLITE,
+            existing_results=provider.ResultStatePolicy.KEEP,
+        ),
+    )
+    request_timeout: int = 125
+    allow_versions: list[str] = field(default_factory=lambda: ["2.0"])
+    # api_key: str = "env:NVD_API_KEY"
+
+    def __post_init__(self) -> None:
```
wagoodman, line 26: can delete the post init since it's a no-op

src/vunnel/providers/mariner/__init__.py
```diff
@@ -0,0 +1,77 @@
...
+        self.logger.debug(f"config: {config}")
+        self.schema = schema.OSSchema()
+        self.parser = Parser(
+            workspace=self.workspace,
+            allow_versions=self.config.allow_versions,
+            download_timeout=self.config.request_timeout,
+            logger=self.logger,
+        )
+
+        if self.config.runtime.skip_if_exists and config.runtime.existing_results != provider.ResultStatePolicy.KEEP:
```
wagoodman, line 57: if this needs to be here then we should be using the common enforcer function here (see https://github.com/anchore/vunnel/blob/bbe25d971eaf537864846494cfe7cce475ca0e00/src/vunnel/providers/wolfi/__init__.py#L47-L48 )

src/vunnel/providers/mariner/__init__.py
```diff
@@ -0,0 +1,77 @@
...
+                "if 'skip_if_exists' is set then 'runtime.existing_results' must be 'keep' "
+                "(otherwise incremental updates will fail)",
+            )
+
+    @classmethod
+    def name(cls) -> str:
+        return "mariner"
+
+    def update(self, last_updated: datetime.datetime | None) -> tuple[list[str], int]:
+        self.logger.info("updating results in Mariner provider")
```
wagoodman, line 68: this log line seems like a duplicate relative to the common statement we have here: https://github.com/anchore/vunnel/blob/bbe25d971eaf537864846494cfe7cce475ca0e00/src/vunnel/cli/cli.py#L151

src/vunnel/providers/mariner/parser.py
```diff
@@ -0,0 +1,196 @@
...
+                Metadata={},  # Ideally, there's no metadata here.
+            )
+
+
+MARINER_URL_BASE = "https://raw.githubusercontent.com/microsoft/CBL-MarinerVulnerabilityData/main/{}"
+MARINER_URL_FILENAME = "cbl-mariner-{}-oval.xml"
+
+
+class Parser:
+    logger = logging.getLogger("mariner-parser")
```
wagoodman, line 167: this line is functional, but should probably be moved to within an instance instead of on the class definition. We've been deriving the logger from the class name if it isn't passed in :
https://github.com/anchore/vunnel/blob/bbe25d971eaf537864846494cfe7cce475ca0e00/src/vunnel/providers/nvd/manager.py#L24-L26

(Also note that you have a logger passed into the constructor but it is currently ignored)

src/vunnel/providers/mariner/__init__.py
```diff
@@ -0,0 +1,77 @@
...
+@dataclass
+class Config:
+    runtime: provider.RuntimeConfig = field(
+        default_factory=lambda: provider.RuntimeConfig(
+            result_store=result.StoreStrategy.SQLITE,
+            existing_results=provider.ResultStatePolicy.KEEP,
+        ),
+    )
+    request_timeout: int = 125
+    allow_versions: list[str] = field(default_factory=lambda: ["2.0"])
```
wagoodman, line 23: should we additionally be processing 1.0 versions as well? though it's being sunset soon, it may still be used by users, thus should be supported on grype scans.

wagoodman: also, this acs more like a allow list instead of a deny list. Ideally we should be able to accept future versions without needing to update code (similar example here https://github.com/anchore/vunnel/blob/bbe25d971eaf537864846494cfe7cce475ca0e00/src/vunnel/providers/rhel/__init__.py#L26). 

Would that be an ok enhancement to this PR?

src/vunnel/providers/mariner/parser.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def __init__(self, workspace: Workspace, download_timeout: int, allow_versions: list[Any], logger: logging.Logger):
+        self.workspace = workspace
+        self.download_timeout = download_timeout
+        self.allow_versions = allow_versions  # TODO: use this to pass in "2.0"?
```
wagoodman, line 172: seems like a stale TODO (looks like this is how this is behaving)

src/vunnel/providers/mariner/__init__.py
wagoodman: actually, I take back some of what I said... we do have inconsistencies here between providers: https://github.com/anchore/vunnel/blob/bbe25d971eaf537864846494cfe7cce475ca0e00/src/vunnel/providers/sles/__init__.py#L24

The SLES provider is acting as an allow list as well.

tests/unit/providers/mariner/test_mariner.py
```diff
@@ -0,0 +1,61 @@
...
+import shutil
+
+import pytest
+from pytest_unordered import unordered
+from vunnel import result, workspace
+from vunnel.providers.mariner import Config, Provider, parser
+from vunnel.providers.mariner.parser import MarinerXmlFile
+from vunnel.utils.vulnerability import Vulnerability, FixedIn
+
+
```
wagoodman, line 13: every provider additionally has a schema-specific test to ensure that the results have the appropriate data shape... can you add that as well?

e.g. https://github.com/anchore/vunnel/blob/bbe25d971eaf537864846494cfe7cce475ca0e00/tests/unit/providers/sles/test_sles.py#L288-L306

src/vunnel/providers/mariner/generate_models.py
```diff
@@ -0,0 +1,47 @@
...
+        w.write(r.content)
+
+
+def main() -> None:
+    versions = ["2.0"]
+    dest_path = tempfile.TemporaryDirectory()
+    for v in versions:
+        download_version(v, dest_path.name)
+
+    we_are_here = os.path.realpath(os.path.dirname(__file__))
```
wagoodman, line 26: you... are... there... 😄

src/vunnel/providers/mariner/model/generated.py
```diff
@@ -0,0 +1,480 @@
+from dataclasses import dataclass, field
+from typing import List, Optional, Union
+from xsdata.models.datatype import XmlDateTime
+
+
+@dataclass
+class Evr:
```
wagoodman, line 7: this is surprisingly clean for generated code ❤️

src/vunnel/providers/mariner/generate_models.py
```diff
@@ -0,0 +1,47 @@
+import os
+import tempfile
+from subprocess import PIPE, Popen
+
+import requests
+
+MARINER_URL_BASE = "https://raw.githubusercontent.com/microsoft/CBL-MarinerVulnerabilityData/main/{}"
+MARINER_URL_FILENAME = "cbl-mariner-{}-oval.xml"
+
+
```
wagoodman, line 10: there should probably be one of the following about what this is and how to use it:

1. a healthy comment here in this file
2. a new README.md or DEVELOPING.md in this `mariner` dir 
3. a note in the top-level DEVELOPING.md

If you're looking for votes, I think I like the `mariner/DEVELOPING.md` option (2) best. (and if you see a better place shout out)

src/vunnel/providers/mariner/parser.py
```diff
@@ -0,0 +1,196 @@
...
+            if not vulnerability_id:
+                continue
+            yield Vulnerability(
+                Name=vulnerability_id,  # intentional; legacy API uses Name as field for vulnerability ID.
+                NamespaceName=self.namespace_name(),
+                Description=self.description(d) or "",
+                Severity=d.metadata.severity,
+                Link=link,
+                CVSS=[],  # NOTE: this is not present in OVAL XML.
+                # This is a list because because there are several versions
```
wagoodman, line 153: nit: this comment seems to be talking about CVSS, if so, it should probably go above that field. I thought while reading this that you were talking about the FixedIn field.

src/vunnel/providers/mariner/parser.py
```diff
@@ -0,0 +1,196 @@
...
+                # Each version is a vector of info about the vulnerability
+                # serialized as a string like CVSS:3.1/AV:N/AC:L/PR:L/UI:N/S:U/C:H/I:H/A:H
+                # TODO: add test image to https://github.com/anchore/test-images
```
wagoodman, line 156: is this TODO current? Also, why add to the test_images repo? (I think I'm missing something is all)

src/vunnel/providers/mariner/parser.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def make_fixed_in(self, definition: Definition) -> FixedIn | None:
+        state = self.get_state(definition)
+        obj = self.get_object(definition)
+        if state is None or state.evr is None:
+            return None
+        if obj is None or obj.name is None:
+            return None
+        version = state.evr.value
+        if state.evr.operation == LTE:
```
wagoodman, line 116: this is fairly tricky, a healthy comment here explaining why `LTE` means we should consider all versions vulnerable would be helpful for future devs here.

src/vunnel/providers/mariner/__init__.py
willmurphyscode: We should definitely allow 1.0 to begin with. It's an allow-list because I haven't come up with a great way to glob on a remove git repo, but agree that a (initially empty) denylist would be better.

src/vunnel/providers/mariner/parser.py
willmurphyscode: Should we just normalize towards "every class makes a logger based in `self.__class__.__name__`? Or are there times when we should be passing down?

src/vunnel/providers/mariner/parser.py
willmurphyscode: Are we ok having the quality tests depend on a digest in a repo we're not hosting? I think I meant that I should build and push an image built from mariner to use in quality tests.

src/vunnel/providers/mariner/parser.py
willmurphyscode: How's this?

> There are 2 choices for state.ever.operation: "less than" or "less than or equal to".
So for each vulnerability, the data either says "versions < 3.2.1 are vulernable"
or the statement "versions <= 3.2.1 are vulnerable". In the second statement,
the data has no information about any fixed version, so we report "None"
as the fixed version, meaning we consider all version vulnerable.
For example, if version 3.2.1 of a library is vulnerable, and is the latest version
mariner data might have "versions <= 3.2.1" is vulnerable. If the library is then fixed in version 3.3.1, the data will be updated to say "versions < 3.3.1", and we can infer that it was fixed in 3.3.1.

src/vunnel/providers/mariner/parser.py
```diff
@@ -0,0 +1,196 @@
...
+
+        self.states_by_id = {}
+        for state_element in etree.XPath("//linux-def:rpminfo_state", namespaces=nsmap)(root):
+            try:
+                element_bytes = etree.tostring(state_element)
+                state_obj = xml_parser.from_string(element_bytes.decode(), RpminfoState)
+                self.states_by_id[state_obj.id] = state_obj
+            except Exception as ex:
+                self.logger.warning(f"skipping rpminfo_object element in {oval_file_path} due to {ex}")
+        self.mariner_version = oval_file_path.split("-")[-2]  # TODO: less brittle way?
```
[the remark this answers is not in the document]
willmurphyscode: Do we need a less brittle way here?

The `TODO` was meant for this PR, but I don't see a way to make this better; if the Mariner XML files break their naming convention, this will break, but I can't think of anything that won't.

src/vunnel/providers/mariner/__init__.py
willmurphyscode: I've added 1.0. Should we add an additional labeled image to the quality tests for 1.0?

src/vunnel/providers/mariner/parser.py
wagoodman: Right now the convention is to pass down from the provider, so everything is logged relative to that provider (and not necessarily the internal objects of that provider, which is an implementation detail and doesn't need to be exposed logging-wise). That being said, in case a logger isn't passed we go ahead at set a default one such that logging doesn't do something like raise an exception.

src/vunnel/providers/mariner/__init__.py
wagoodman: > Should we add an additional labeled image to the quality tests for 1.0?

if there is a difference in how it's processed, yes. Otherwise there isn't a need to add it. (it looks like it's the same, so no need to add labels for it).

src/vunnel/providers/mariner/parser.py
wagoodman: > Are we ok having the quality tests depend on a digest in a repo we're not hosting

indeed, we already have precedence for this too. So no need to capture a separate image in `test_images` unless you're trying to make the image / label set smaller.

src/vunnel/providers/mariner/parser.py
willmurphyscode: Thanks! I'll push a rev that passes down the logger.

src/vunnel/providers/mariner/__init__.py
willmurphyscode: Thanks, I'll remove this `if` check in a rev.

Is 

```
existing_results=provider.ResultStatePolicy.KEEP
```

still correct?

wagoodman: yup, yup 👍

wagoodman (approved):
nice addition 😎 I'm OK as-is, but understand if you want to add further adjustments based off of https://github.com/anchore/vunnel/pull/181#issuecomment-1561780921 + any other discussion from folks on the team.

spiffcs:
> The update strategy - in this draft PR, the downloader just always runs. We need to figure out a better strategy there.

Definitely think this can be a future PR thing - current PR looks good and it's fine for now that it always runs

> The XML we observe over at https://github.com/microsoft/CBL-MarinerVulnerabilityData is narrower than the OVAL XML specification allows. (For example, every criteria node seems to have AND as its operator.) We might need to talk about future proofing. The current revisions future proofing is basically done by parsing the XML one CVE at a time and logging failures.

Nothing to add beyond what was discussed

> This uses xsdata, a new dependency, to generate dataclasses representing the OVAL XML found in the repo

Nice change and really appreciated the FAQ!

Whole PR is LGTM - I don't have significant comments to add that would halt this or change the discussion from earlier comments. My biggest concern might be the update strategy, but that's more of a wait and see if it needs any more fine tuning beyond always running.

spiffcs (approved):

PR #184, opened 2023-05-19
remove the long deprecated "centos" provider

westonsteimel:
The "centos" provider was only originally included as the RedHat provider relied on it, but in actuality it was always RedHat data.  This removes the provider logic and moves the necessary oval parsing logic directly into the `rhel` provider.


wagoodman:
also need to remove https://github.com/anchore/vunnel/blob/v0.10.0/tests/quality/config.yaml#L124

PR #199, opened 2023-05-29
Draft: feat: add osv schema and rocky provider by downloading google bucket

jneate:
Add Android as a new provider by downloading the Android vulns from OSV's google bucket.

Raising the PR to check the change is wanted as I can see on slack there's still questions about how the integration with osv.dev will work going forwards, the schema is lifted from the osv-schema repo and returned back into the DB. I'm happy to do some work towards handling that schema in the other components but figured a discussion was probably needed.

_Note for maintainers_: Be sure to add the `run-pr-quality-gate` label to the PR when it has stabilized.


wagoodman:
Your code contribution looks good, however, we've been looking at the upstream data quality and are uncertain if this data source makes sense relative to the upstream canonical data source (from rocky linux directly). A few questions came up:

1. Severity is not captured. OSV allows for capturing CVSS v2 or v3 severity but this doesn't appear to be populated in the OSV data. Upstream this is captured as a category string (e.g. "critical" or "low").
2. RPM modularity might not be accounted for (example fix for existing RPM-based providers https://github.com/anchore/vunnel/pull/36). I've been looking for examples of RPM modules being listed on records without success yet. It's possible that RPM modules are not leveraged in Rocky linux, however, we'd need to make certain this is the case before proceeding since this has been a source of false positives in the past for other RPM-based ecosystems.

I think both of these elements could be included in records upstream in the [`database_specific`](https://ossf.github.io/osv-schema/) fields for the affected records. It looks like the OSV data for Rocky Linux is relatively new (https://groups.google.com/g/osv-discuss/c/lWurEpuSMOc) so it's possible that these details were overlooked and an adjustment upstream needs to be made (or that there are specific reasons why this information was not included).

Unfortunately what this means for this PR is that we won't be able to continue until these questions are answered or addressed upstream.

wagoodman:
I'm going to close this PR for now, but shout out if there is additional progress / ideas that you have here and we can re-open or chat over an issue.

PR #214, opened 2023-06-23
fix: handle more ubuntu git repo errors

westonsteimel:
Currently there could be situations where there are untracked files present in the ubuntu git working directory (likely due to some previous failure state), so the provider will now automatically issue a `git clean -fd` to try and resolve these rather than failing.  If the pull still fails, then just re-clone the repo (except in the case of 503 errors)

Now raises a specific exception for the 503 errors which are frequently issued by the upstream https endpoint so that specific scenario can be handled separately


src/vunnel/providers/ubuntu/git.py
```diff
@@ -358,7 +389,12 @@ def _exec_cmd(self, cmd, *args, **kwargs) -> bytes:
         try:
             self.logger.trace("running: {}".format(cmd))
             cmd_list = shlex.split(cmd)
-            return subprocess.check_output(cmd_list, *args, **kwargs)  # nosec
+            return subprocess.check_output(cmd_list, *args, **kwargs, stderr=subprocess.PIPE)  # nosec
         except Exception as e:
             self.logger.exception("error executing command: {}".format(cmd))
+
+            if isinstance(e, subprocess.CalledProcessError):
+                if e.stderr and self._ubuntu_server_503_message in e.stderr.decode():
```
kzantow, line 397: nit: this seems a bit brittle

kzantow (approved):
👍

PR #223, opened 2023-06-29
Update os in release workflow

wagoodman:
This makes the OS version consistent in the validations and release pipelines (important when sharing a cache between workflows). I can't seem to find a way to get the `ubuntu-20.04` name/version info from the github runner context (seems to not exist https://github.com/actions/runner/issues/2527 ) but that would be ideal to bake into the cache key as well.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -9,7 +9,7 @@ on:
 jobs:
 
   quality-gate:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v3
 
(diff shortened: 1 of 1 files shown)


westonsteimel (approved):

kolyshkin:
> I can't seem to find a way to get the `ubuntu-20.04` name/version info from the github runner context (seems to not exist [actions/runner#2527](https://github.com/actions/runner/issues/2527) ) but that would be ideal to bake into the cache key as well.

There is apparently `${process.env.ImageOS}` (see https://github.com/actions/setup-go/pull/383), which resolves to e.g. `ubuntu20` or `ubuntu22` (which, I guess, is OK since GHA only support LTS releases).

PR #232, opened 2023-07-10
feat: preserve prior results longer with sqlite writer

westonsteimel:
When using the sqlite writer, preserve the prior results as long as possible by writing the new results to a temp file and then renaming it after everything is successful.

### DIFF
--- a/src/vunnel/result.py
+++ b/src/vunnel/result.py
@@ -4,6 +4,7 @@ import abc
 import enum
 import logging
 import os
+import shutil
 import time
 from dataclasses import asdict, dataclass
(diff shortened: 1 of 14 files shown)


src/vunnel/result.py
```diff
@@ -155,28 +184,35 @@ def close(self) -> None:
             self.engine = None
             self.table = None
 
+        if successful:
+            os.rename(self.temp_db_file_path, self.db_file_path)
+        elif os.path.exists(self.temp_db_file_path):
```
wagoodman, line 189: can you do this in prepare as well (with a warning?)

wagoodman (approved):

PR #244, opened 2023-07-18
Fix grype1393

willmurphyscode:
Fixes https://github.com/anchore/grype/issues/1393

Fix Mariner provider to stop including CVEs that are "Not Applicable." Additionally, fix the data retention to regenerate the Mariner database, rather than appending to it, since we are downloading entire OVAL files every time.


kzantow (approved):
LGTM

PR #269, opened 2023-08-29
fix: always include rhel out of support entries when no other fix info available for any other releases

westonsteimel:
Out of support entries from the RHEL provider should always be included as affected if there is no other fixed-in or affected information from other releases


wagoodman (approved):

westonsteimel:
superseded by #270

Issue #294, opened 2023-09-21
Add GitHub Security Advisory data for Dart packages


westonsteimel:
**What would you like to be added**:
Add support for pulling in [GitHub Security Advisory data for Dart](https://github.com/advisories?query=type%3Areviewed+ecosystem%3Apub)

**Why is this needed**:
Syft supports cataloging of Dart packages but there is currently no vuln data being built into grype-db to match against in grype

**Additional context**:
Should hopefully just be a simple add to the mapping in: https://github.com/anchore/vunnel/blob/cd2f18eaaa403db90595b75eb38c4390b6de7525/src/vunnel/providers/github/parser.py#L32-L41

PR #301, opened 2023-09-27
feat: fetch GHSA data for dart, swift, and erlang

### LINKED ISSUE #300: Add GitHub Security Advisory data for Erlang packages
Add support for pulling in [GitHub Security Advisory data for Erlang](https://github.com/advisories?query=type%3Areviewed+ecosystem%3Aerlang)

**Why is this needed**:
Syft supports cataloging of Erlang packages but there is currently no vuln data being built into grype-db to match against in grype

**Additional context**:
Should hopefully just be a simple add to the mapping in: https://github.com/anchore/vunnel/blob/cd2f18eaaa403db90595b75eb38c4390b6de7525/src/vunnel/providers/github/parser.py#L32-L41

westonsteimel:
Adds support for fetching GitHub Security Advisories for:

- dart
- swift
- erlang

Closes #300 #294 #293

### DIFF
--- a/src/vunnel/providers/github/parser.py
+++ b/src/vunnel/providers/github/parser.py
@@ -30,14 +30,17 @@ from vunnel.utils import fdb as db
 from vunnel.utils.vulnerability import CVSS, CVSSBaseMetrics
 
 ecosystem_map = {
+    "ERLANG": "erlang",
     "COMPOSER": "composer",
     "GO": "go",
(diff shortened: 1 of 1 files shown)


westonsteimel:
Erlang is a bit more complicated to support, so breaking off the others into https://github.com/anchore/vunnel/pull/302 and leaving Erlang for a future discussion

PR #316, opened 2023-10-04
Update vunnel list output in README.md

andrew:
I found centos being listed in the readme but not in my output a bit confusing, so updated the list to match what I got today.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,9 +52,10 @@ $ vunnel list
 
 alpine
 amazon
-centos
+chainguard
 debian
 github
+mariner
 nvd
 oracle
 rhel


andrew:
Side note: would be good to turn on the admin option for requiring signoff for web commits to avoid having to do the clone/rebase dance for minor changes made via the github web ui: https://github.blog/changelog/2022-06-08-admins-can-require-sign-off-on-web-based-commits/

spiffcs (approved):

spiffcs:
Thanks for the suggestion and PR @andrew - let me see if I can get that option enabled - apologies for the friction here

Issue #310, opened 2023-09-29
Disallow all bare try-except clauses


wagoodman:
Today we have several [`noqa: E722`](https://www.flake8rules.com/rules/E722.html) statements as well as ignoring blank `noqa` statements ([here](https://github.com/anchore/vunnel/blob/16efe8a426ca0716a8eacf8ed09ca56410825263/pyproject.toml#L218) and [here](https://github.com/anchore/vunnel/blob/16efe8a426ca0716a8eacf8ed09ca56410825263/pyproject.toml#L221)). We should:
- not allow bare noqa statements, as this can hide a lot of issues
- not allow any E722 exceptions, as this is a popular antipattern in this codebase and has lead to production regressions

Additionally, all `except:` blocks should log the exception and some useful context, to make feed failures easier to diagnose from logs.

willmurphyscode:
All the refactors we planned under this issue to make it easier / safer to work in Vunnel are implemented.

PR #361, opened 2023-10-24
chore: remove file noqa from rhel oval parser

willmurphyscode:
This way, changes to this file will still get linted.

See #310


spiffcs (approved):
Nice to get these files down to the direct issues that we can clean up eventually - thanks for the PR!

PR #373, opened 2023-10-27
chore: remove file noqa from utils fdb

willmurphyscode:
Apparently this noqa wasn't hiding anything.

See #310


willmurphyscode:
It looks like this failed with `2023-10-27T13:19:31.2300897Z     version = response.json()["name"]` throwing a KeyError at `yardstick/tool/syft.py line 139`. This has passed on a retry before; possibly a temporary issue pulling the latest version of a release.

I'll retry, but also bring it to the team's attention in case something is off here.

wagoodman (approved):

PR #376, opened 2023-10-30
refactor: introduce http get utility; use it in mariner provider

willmurphyscode:
Create a wrapper for `requests.get` at utils/http.py. The wrapper should always log the URL on failures, and should handle retries. Unit test the wrapper, and use it in the mariner provider.

Follow up PRs will introduce the wrapper into other providers.

See #310


src/vunnel/providers/amazon/parser.py
```diff
@@ -100,7 +100,7 @@ def _get_alas_html(self, alas_url, alas_file, skip_if_exists=True):
             return content  # noqa: RET504
 
         try:
-            self.logger.debug(f"downloading ALAS from {alas_url}")
+            self.logger.info(f"downloading ALAS from {alas_url}")
```
wagoodman, line 103: why change this to info? Since this downloads a lot of info and can get rather chatty, this feels like a lower logging level.

wagoodman: same comment in a few places

wagoodman (commented):
There might be an opportunity for refactoring downloading files from a single util function that we can reuse everywhere. That way logging is done the same way, at the same level, can be stubbed out the same way, etc.

src/vunnel/providers/amazon/parser.py
wagoodman: based on an offline conversation -- ignore this

wagoodman (approved):
we can do the shared util function in the future

src/vunnel/utils/http.py
```diff
@@ -0,0 +1,45 @@
...
+
+def get(
+    url: str,
+    logger: logging.Logger,
+    retries: int = 5,
+    backoff_in_seconds: int = 3,
+    timeout: int = DEFAULT_TIMEOUT,
+    **kwargs: Any,
+) -> requests.Response:
+    logger.debug(f"attempting get request on {url}")
```
wagoodman, line 18: nit/suggestion: since this will show up a lot, the smaller the better
```suggestion
    logger.debug(f"http GET {url}")
```

src/vunnel/utils/http.py
```diff
@@ -0,0 +1,45 @@
...
+            will_retry = ""
+            if attempt + 1 < retries:
+                will_retry = f" (will retry in {backoff_in_seconds} seconds) "
+            # this is an unexpected exception type, so include the attempted request in case the
+            # message from the unexpected exception doesn't.
+            logger.warning(f"attempt {attempt + 1} of {retries}{will_retry}: unexpected exception during GET {url}: {e}")
+    if isinstance(last_exception, BaseException):
+        logger.error(f"last retry of GET {url} failed with {last_exception}")
+        raise last_exception
+    raise Exception("unreachable")
```
wagoodman, line 45: why not this?

```suggestion
    logger.error(f"last retry of GET {url} failed with {last_exception}")
    raise last_exception
```

tests/unit/providers/mariner/test_mariner.py
```diff
@@ -91,7 +91,7 @@ def disable_get_requests(monkeypatch):
     def disabled(*args, **kwargs):
         raise RuntimeError("requests disabled but HTTP GET attempted")
 
-    monkeypatch.setattr(parser.requests, "get", disabled)
+    monkeypatch.setattr(utils.http, "get", disabled)
```
wagoodman, line 94: This could be something like this:
```python
@pytest.fixture()
def disable_get_requests(monkeypatch):
    def disabled(*args, **kwargs):
        raise RuntimeError("requests disabled but HTTP GET attempted")
    from vunnel import utils
    monkeypatch.setattr(utils.http, "get", disabled)
```

and put into `conftest.py` generically for anyone to use

wagoodman (approved):
🙌

PR #385, opened 2023-11-01
refactor: use new http wrapper in nvd provider

willmurphyscode:
Use the http wrapper from #376 in the NVD provider.

See also #310.


src/vunnel/providers/nvd/api.py
```diff
@@ -119,7 +120,7 @@ def _request_all_pages(
 
         payload = orjson.loads(response.text)
         if "message" in payload:
-            raise RuntimeError(f"API error: {response['message']}")
+            raise RuntimeError(f"API error: {payload['message']}")
```
willmurphyscode, line 123: This change is not logically part of this PR, but the standard library `requests.get` doesn't annotate a return type, while the wrapper guarantees that it returns a `requests.Response`. Once the `requests.Response` was guaranteed to be the type of response on line 119, the static analysis tools can tell the indexing into response wasn't going to work.

Based on the test `if "message" in payload:`, I'm assuming that the raise was meant to take the message from the payload.

I'm open to putting this in its own PR if anyone feels strongly about it.

wagoodman: this looks like a good catch

wagoodman (approved):

PR #389, opened 2023-11-08
Switch from dataclass-wizard to mashumaro

wagoodman:
This fixes the [issues we're seeing with dataclass-wizard](https://github.com/rnag/dataclass-wizard/issues/89) by swapping it out with [mashumaro](https://github.com/Fatal1ty/mashumaro). This additionally swaps the CI validations to a matrix to test multiple versions of python. This was done by adding `tox` back into the project. Nothing has changed from the `make unit` workflow, however if you want to test with all supported versions of python (and you have already installed them locally) then you can run `make unit-matrix`. CI runs with this matrix configuration in mind.

Note: this updates the default python version in CI to python 3.11. This depends on the defaults in the bootstrap action locally in this repo. SInce this is not a single source of truth, we're at risk of drift here (relative to the python versions in the test matrix).

TODO:
- [ ] ensure CI is passing, and each python version in each job is correct (in the matrix run)
- [ ] passing quality gate: ensure all providers are tested


westonsteimel:
Hmm, something is off with the unit tests.  They're all running on python 3.10

westonsteimel:
Ah, looks like something weird happening in the poetry install: https://github.com/anchore/vunnel/actions/runs/6801136283/job/18491300544?pr=389#step:3:103

westonsteimel:
installing it in python3.10

westonsteimel:
Hmm, I'm not really sure why it is only installing poetry in python3.10.  I tried bumping to a newer poetry version but that did nothing

wagoodman:
I've removed the `tox-gh-actions` tool that I was trying to use to facilitate the matrix seamlessly. The trade off is adding some seams: the makefile is not the entrypoint anymore for running tox.

zhill:
I suggest looking at the tox allowlist section. it's using poetry from the initial install I think rather than from the tox matrix.

willmurphyscode (approved):

PR #392, opened 2023-11-09
refactor: new http wrapper in rhel provider

willmurphyscode:
This uses the new HTTP GET wrapper in the RHEL provider. It also adds a new pattern, where the client can pass in a `status_handler` that will be called instead of `response.raise_for_status` to determine whether the request succeeded. This is because the RHEL provider doesn't retry 404s, so some inversion of control was necessary to preserver it's behavior. I'm looking for feedback on the approach.


zhill:
Hi @willmurphyscode !  This is an interesting question but I think we can look at the abstractions expected from each function to guide us.

It feels to me like this is an acceptable case where the caller should catch the response.raise_for_status error and handle the 404 in a specific way since that is an error. Only the caller knows that an error (from the perspective of the http lib) is part of an expected path, so exception handlers seem justified there rather than adding complexity into the http layer itself. For example, the RHEL provider can wrap the http call and that keeps the RHEL path clean and also keeps the HTTP layer simple and http-focused.

Just my 2c.

willmurphyscode:
@zhill I agree, but there would still be a subtle difference in the behavior: how does the caller tell the http.get wrapper not to retry 404s? (Maybe this is an unimportant feature of the current implementation, but right now it retries every status code _except_ 404 and 200, while the http.get wrapper will retry 404s in an unconfigurable way.) I also thought about just passing in a collection of status codes that shouldn't be retried, but that might be too much less general.

zhill:
Interesting 🤔 I wouldn't expect a http-level client to retry 4xx in general. I think the expectation for those errors is that they are not expected to work on a retry because the client is sending a bad request of some sort. I suspect that you're seeing API-specific logic here where, for example, the API may use 403s for rate-limit backoffs. That behavior is specific to the API that this provider is using, so I wouldn't bake it into the http utility itself, but the RHEL-data client layer.

I believe the goal of consolidation here is to unify client config and visibility (proxies, timeouts, logging etc) rather than response behaviors so I think what I'm suggesting is consistent with that goal.

wagoodman (approved):
@zhill I agree that we probably shouldn't retry on 4xx, I think we can refactor all of the providers to do just that (with some exceptions) in another PR. This at least gets these code paths to a single choke point to help with the refactor and make the change globally (allowing for inverting the control for error conditions back to the caller).

PR #397, opened 2023-11-13
fix: NVD api key should not be required

westonsteimel:
Although providing an API key for the NVD provider is highly encouraged, it is not required and is not set in all of our environments.

### DIFF
--- a/src/vunnel/providers/nvd/__init__.py
+++ b/src/vunnel/providers/nvd/__init__.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import os
 from dataclasses import dataclass, field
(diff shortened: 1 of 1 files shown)


src/vunnel/providers/nvd/__init__.py
```diff
@@ -20,10 +20,10 @@ class Config:
         ),
     )
     request_timeout: int = 125
-    api_key: str = "env:NVD_API_KEY"
+    api_key: str | None = "env:NVD_API_KEY"
```
willmurphyscode, line 23: The think python 3.9 wants the `Optional[str]` syntax.

willmurphyscode (approved):

PR #399, opened 2023-11-15
fix(wolfi): remove filtering on vuln ids

westonsteimel:
This removes the filtering to allow all specified vulnerability ids to flow through from the wolfi advisory feed


luhring:
Wow, this is awesome! Will this fix the case for `chainguard` as well?

kzantow (approved):

westonsteimel:
Ah, @luhring, I think you're correct that this should also address the chainguard ones since it is just reusing the wolfi provider

PR #407, opened 2023-11-27
chore: replace usage of json with orjson

westonsteimel:
Replaces all usage of the built-in `json` library with `orjson`.  We were already using it in a couple of places within vunnel and have found some significant memory and speed improvements from switching to it elsewhere.  This also allows eliminating the `rfc3339` python package as a prod dependency and it is now only brought in as part of yardstick.  This is because `orjson` already converts datetimes to rfc3339 format by default


wagoodman (approved):
do we have a linter rule we could add to prevent regression? (say, fail if we find `json.load*` or `json.dump*`)

westonsteimel:
> do we have a linter rule we could add to prevent regression? (say, fail if we find `json.load*` or `json.dump*`)

Yeah, that's what the [tool.ruff.lint.flake8-tidy-imports.banned-api] section in pyproject.yaml does

PR #417, opened 2023-12-08
Feat/data enhancement spike

willmurphyscode:
draft for getting feedback


transforms/github/changes.yaml
```diff
@@ -0,0 +1,8 @@
+# pretend the following PRs were merged
+include-prs:
+  - https://github.com/github/advisory-database/pull/2630
+
+# assuming the following GHSAs were reviewed, even if they
+# are not returned by the API
+assume-reviewed:
```
willmurphyscode, line 7: From Josh: This mechanism probably isn't needed. If we're going to include an unreviewed GHSA, we're going to have to change it anyway. So just have a mechanism for changing fetched GHSA data, and assume any changed GHSA should be included in our output, regardless of whether GHSA has reviewed it.

transforms/github/GHSA-prp9-9gxw-38j8.json
```diff
@@ -0,0 +1,73 @@
+{
```
willmurphyscode, line 1: This could also be in GHSA's graphQL or rest API format.

willmurphyscode:
From Josh: For fixing the severity of CVE-2023-44487 in debian namespaces, we could solve the immediate problem by updating it to "high".

willmurphyscode:
this hasn't been touched in years. I think other work has replaced it.

PR #418, opened 2023-12-08
fix: backoff and jitter in new http wrapper

### LINKED ISSUE #415: http.get does not support exponential backoff on retries
**What happened**:

[retry_with_backoff](https://github.com/anchore/vunnel/blob/b4559ae59c190e14db84d4eaa0ef8db21ec52e57/src/vunnel/utils/__init__.py#L34) did an exponential backoff on retries, but the newer https://github.com/anchore/vunnel/blob/main/src/vunnel/utils/http.py#L11 just does a static wait between retries.  The exponential backoff is important for providers like NVD

**What you expected to happen**:

http.get supports an exponential backoff and does it by default

willmurphyscode:
This behavior was present in the old decorator, and is needed for some providers.

Fixes #415.

Also fixes fence post bug in logging that made the wrapper appear to make an extra request.

### DIFF
--- a/src/vunnel/utils/http.py
+++ b/src/vunnel/utils/http.py
@@ -1,4 +1,5 @@
 import logging
+import random
 import time
 from collections.abc import Callable
 from typing import Any, Optional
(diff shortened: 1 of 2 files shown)


wagoodman (approved):

westonsteimel:
Thanks Will!

PR #424, opened 2023-12-20
fix: correct GHSA URLs

willmurphyscode:
Previously, all Wolfi vulnerabilities handled by Vunnel were from NVD CVEs. However, now that Wofli also reports GHSA vulnerabilities as well, switch the link URL on the vulnerability prefix, so that we output the correct URLs.


willmurphyscode:
Closed in favor of https://github.com/anchore/vunnel/pull/425, which fixes additional related URL problems and fixes the quality gate as well.

PR #433, opened 2024-01-02
Move away from asset promotion

wagoodman:
One consequence of auto merging dependabot PRs is that releases cannot depend on workflows always being run on main. This is because when dependabot merges a PR any "recursive" workflows are not kicked off since the default GITHUB_TOKEN was used (a security feature of github actions). We still want the benefit of not adding our own token and still getting dependabot PRs merged automatically, so the next choice we have is to move away from asset promotion and instead create all assets at release.


willmurphyscode (approved):

wagoodman:
We've decided to not go this direction after all

PR #434, opened 2024-01-02
Check PR author login, instead of actor

willmurphyscode:
Otherwise, the workflow won't run on events where another user modified a dependabot PR.


wagoodman (approved):
really subtle

PR #477, opened 2024-02-06
fix(ubuntu): emit vuln rows for out of support

westonsteimel:
Ensure the ubuntu provider emits `wont-fix` vulnerability rows when the final state ends up with a status of `ignored` and matches one of the known end of support labels, and there was no previous fix found after traversing git state


wagoodman (approved):

westonsteimel:
This will address the false negatives reported in https://github.com/anchore/grype/issues/1695

PR #505, opened 2024-03-08
fix: remove sha256 verification for oval parser

spiffcs:
## Summary
This removes the sha256 check from the oval provider so that users will always skip the checksum matching when parsing the following oval data.

It was previously used as a method to skip the download if there was no change. We now opt to always download and remove the complexity based on the below manifest not always having a sha to check against

https://www.redhat.com/security/data/oval/v2/PULP_MANIFEST

### DIFF
--- a/src/vunnel/providers/rhel/oval_parser.py
+++ b/src/vunnel/providers/rhel/oval_parser.py
@@ -1,10 +1,8 @@
 from __future__ import annotations
 
 import bz2
-import hashlib
 import logging
 import os
-import re
 
 import requests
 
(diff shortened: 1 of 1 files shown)


spiffcs:
https://github.com/anchore/vunnel/actions/runs/8198267820/job/22421529479?pr=505 <-- This run shows without the configuration change

I'm going to go back and edit this PR with the config change that allows us to skip the checksums

spiffcs:
@wagoodman question for the am sync:

I wanted this to be secure by default. The default option right now is to NOT skip checksums as you can see by the failed quality gate.

I want to update the test/quality here:
https://github.com/anchore/vunnel/blob/1416dd14345b65ef14ddf24859f2c7da481080e7/tests/quality/configure.py#L598-L601 

This section of the test code is where we execute the different providers for the quality gate.

I want to pass a vunnel config with the updated values explicitly skipping the checksum check for rhel so we have a spot to go back to and take out when we get an answer from Redhat.

Do you think this is the right spot or is there another spot we should be looking to pass this new configuration to generate the db?

When grype-db consumes this change where do you think the best place is to put the new non default vunnel config?

westonsteimel:
If I remember correctly the digest isn't even being used for any sort of security checks, it's just being used as a way to skip re-downloading content if the file hasn't changed from what is already on disk, but I don't think that complexity is worth it really.

westonsteimel:
So I'd propose just making this the default and then removing all of the logic later if we don't just want to do that now

wagoodman (approved):

PR #447, opened 2024-01-09
feat: add support for OSV schema

juan131:
# Summary

As described at https://github.com/anchore/grype/issues/1609, Bitnami is providing vulnerability matching data for their containers via the [Bitnami Vulnerability Database repository](https://github.com/bitnami/vulndb).

This database uses the [Open Source Vulnerability format](https://ossf.github.io/osv-schema) (AKA OSV) and it's currently part of the [aggregated OSV vulnerability database](https://osv.dev/list).

As a first step towards consuming this CVE feed and improve Grype scanning capabilities on Bitnami containers, this PR adds support for the OSV schema to vunnel.

PR #512, opened 2024-03-12
feat: add Bitnami as new provider

juan131:
# Summary

As described at https://github.com/anchore/grype/issues/1609, Bitnami is providing vulnerability matching data for their containers via the [Bitnami Vulnerability Database repository](https://github.com/bitnami/vulndb).

This PR follows up https://github.com/anchore/vunnel/pull/447 adding support for this new provider. This database is imported by cloning the Bitnami Vulnerability Database repository and loading/normalizing all the data from it.


willmurphyscode:
Hi @juan131, sorry that this has sat for a while. I'll see about getting some of this static analysis and unit tests passing today and see if some of us can take a look soon at getting this reviewed and merged.

juan131:
@willmurphyscode I addressed all the linter warns reported by `poetry run make lint-fix` but the static analysis is failing and I'm not 100% sure the reason since it's complaining about pieces of code that are copy-pasted from other providers

carrodher:
Hi @wagoodman @willmurphyscode, is there anything we can help with at this time?

willmurphyscode:
Hi @carrodher and @juan131! Thanks for the PR!

> is there anything we can help with at this time?

Do you have an image or Dockerfile that you'd expect to exhibit vulnerabilities pulled in from this feed?

The best end-to-end test here would be to find/make an image that has vulnerabilities reported by the new feed, and then bake a new grype database from this branch and make sure that when grype uses the new database, it finds the new vulnerability.

We are happy to do some of the boring stuff (like making linters pass) but it would be super helpful if you all could find/make an image or Dockerfile for us to use in an end-to-end test.

juan131:
Hi @willmurphyscode 

Every Bitnami container image contains one or multiple SPDX files containing the containers BOM. The vulnerabilities associated to the packages listed on these SPDX files can be found in the Bitnami Vulnerability Database. This is explained in the link below:

- https://github.com/bitnami/vulndb?tab=readme-ov-file#how-to-consume-this-cve-feed

willmurphyscode:
After going some more looking, it seems like this is blocked on getting anchore/grype-db#217 merged, and adding a transformer for the new OSV schema. (Without that work, we can't make a grype database that includes the Bitnami data, so we can't run an end-to-end test, so we can't know that this PR is really working.)

After that, we can build a test image to add end to end tests to this PR, clean up linting, and then get this merged.

I'm going to take a look at anchore/grype-db#217 first.

src/vunnel/providers/bitnami/parser.py
```diff
@@ -0,0 +1,101 @@
...
+                        "NamespaceName": namespace,
+                        "Version": version,
+                    },
+                )
+        link = "None"
+        if "references" in vuln_entry and len(vuln_entry["references"]) > 0:
+            link = vuln_entry["references"][0]
+
+        return vuln_id, {
+            "Vulnerability": {
```
willmurphyscode, line 83: I think we want the second return value here to conform to the OSV schema.

I'm simplifying a bit, but second element of the returned tuple will be JSON-serialized and eventually picked up by the anchore/grype-db#217 (and some follow up work) and transformed into the Grype-specific vulnerability schema that Grype matches on.

This dict seems to imitate the shape of some dictionaries returned by other parsers, which looks right, but in this case I think we want a real OSV record.

I'm going to try to do some refactoring to make it more obvious what things `parser.get()` should return.

willmurphyscode:
So I was wrong about anchore/grype-db#217 being the first step. I think the first step is really anchore/syft#3065. In other words, the tools need detect the Bitnami packages and encode them in the SBOM in a consistent way before we can match vulnerabilities against them. I'm sorry to hold this up further, but I think the right next step is to figure out answer to the couple of design questions raised in anchore/syft#3065.

Thanks for your patience!

juan131:
Hi @willmurphyscode 

Thanks for looking into this.

> the tools need detect the Bitnami packages and encode them in the SBOM

Why is this step required? I mean, Bitnami is already doing this job and ensuring every package is properly described in the SPDX files. AFAIK, the tools should only have capabilities to detect the `.spdx` files included by Bitnami in the filesystem & recognize them. Maybe I'm missing sth.

Regarding the concern you mentioned at https://github.com/anchore/syft/issues/3065 about duplicates, a mechanism to detect and remove them should be included (e.g. based on the pURL).

willmurphyscode:
@juan131 the reason for the Syft change is that, by default, the SBOM cataloger is off (that is, Syft doesn't incorporate SBOMs it finds in images into the SBOM it produces by default. It used to, but we found this confused people). But I think when we're scanning a Bitnami image, we want to respect the SPDX SBOM we find and raise up those packages. We can discuss details over at anchore/syft#3065.

carrodher:
Hi @willmurphyscode,

I hope you’re doing well! We wanted to check in on the status of our recent contributions. We have some open PRs, and we’re unsure about the current progress or the next steps. Could you kindly provide an update? It would help us plan our work and understand how we can continue contributing effectively.

Here are a couple of the open PRs for reference:
* https://github.com/anchore/grype-db/pull/217
* https://github.com/anchore/grype/pull/1826

We appreciate any details you can share. Thanks in advance for your help!

willmurphyscode:
> @willmurphyscode I addressed all the linter warns reported by poetry run make lint-fix but the static analysis is failing and I'm not 100% sure the reason since it's complaining about pieces of code that are copy-pasted from other providers

@juan131 this just happened to me, and I learned why, so I thought I would share! 

https://github.com/anchore/vunnel/blob/697e8a2afc748e65a3852566c4e41ff5588427b2/pyproject.toml#L105-L124

This disables `mypy` on some files by path (with no marking in the file, because that's how `mypy` works), so if you take a file from one of those paths and copy it to a different path, static analysis will start failing. I agree that this isn't an ideal situation, and we'll fix it as we have time. But at least it's not a mystery now!

juan131:
Thanks for sharing @willmurphyscode

src/vunnel/providers/bitnami/parser.py
juan131: Addressed at https://github.com/anchore/vunnel/pull/512/commits/333564d6ee41d27a1972db5c60718dc4027d4db4
Please let me know if this is what you meant

src/vunnel/providers/bitnami/parser.py
```diff
@@ -0,0 +1,67 @@
...
+                    yield orjson.loads(f.read())
+
+    def _normalize(self, vuln_entry: dict[str, Any]) -> tuple[str, dict[str, Any]]:
+        self.logger.trace("normalizing vulnerability data")  # type: ignore[attr-defined]
+
+        # We want to return the OSV record as it is (using OSV schema)
+        # We'll transform it into the Grype-specific vulnerability schema
+        # on grype-db
+        vuln_id = vuln_entry["id"]
+        return vuln_id, vuln_entry
```
wagoodman, line 57: this should probably read the schema value from the record and return it up to the Provider class on each record

src/vunnel/providers/bitnami/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+        return "bitnami"
+
+    def update(self, last_updated: datetime.datetime | None) -> tuple[list[str], int]:
+        # TODO: use of last_updated as NVD provider does to avoid downloading all
+        # vulnerability data from the source and make incremental updates instead
+        with self.results_writer() as writer:
+            for vuln_id, record in self.parser.get():
+                writer.write(
+                    identifier=vuln_id.lower(),
+                    schema=self.schema,
```
wagoodman, line 57: we should honor the schema version that is in the record itself, and if the major version differs from that of what is in `__schema__` on the class we should fail

willmurphyscode (approved):

juan131:
Thanks so much for adding the missing pieces, address the linter issues and add the "brotli" test, really appreciated!

PR #519, opened 2024-03-18
fix: more verbose ubuntu provider logging to clearly show progress

zhill:
Moves some debug messages to info level. This does make the logs much more verbose, but for cases where the task is long-running, its helpful to be able to verify that progress is being made.

I'm open to other approaches such as chunked-counters if this is too verbose. My goal is only to ensure that when this provider runs for many minutes to hours, that its clear in the logs that it is working and making progress rather than being frozen at info-level logs.


willmurphyscode:
@zhill this seems to be irrelevant to the new ubuntu parser path, so I'm closing it. Feel free to re-open or just let me know if this was still needed.

PR #520, opened 2024-03-19
Feat/fetch cached workspace

willmurphyscode:


src/vunnel/provider.py
```diff
@@ -136,6 +166,41 @@ def _update(self) -> None:
...
+
+    def _fetch_listing_document(self) -> distribution.ListingDocument:
+        url = self.runtime_cfg.import_url(provider_name=self.name())
+        resp = http.get(url, logger=self.logger)
+        resp.raise_for_status()
+
+        return distribution.ListingDocument.from_dict(resp.json())
+
+    def _has_newer_archive(self, latest_entry: distribution.ListingEntry) -> bool:
+        # TODO: can do checksum comparison here
```
willmurphyscode, line 188: @wagoodman I'm not sure what checksum comparison you mean here. Do you mean making sure the listing entry passes its own checksum validation and falling back to the next oldest if it doesnt?

src/vunnel/provider.py
```diff
@@ -212,3 +277,33 @@ def results_writer(self, **kwargs: Any) -> result.Writer:
...
+
+    # download the URL for the archive
+    resp = http.get(entry.url, logger=logger, stream=True)
+    resp.raise_for_status()
+    logger.debug(f"downloading {entry.url} to {archive_path}")
+    with open(archive_path, "wb") as fp:
+        for chunk in resp.iter_content(chunk_size=None):
+            fp.write(chunk)
+
+    # TODO: ensure the checksum matches whats in the listing entry
```
willmurphyscode, line 293: @wagoodman is additional validation still needed besides `temp_ws.validate_checksums()` above? Or is this `TODO` done?

wagoodman: this is a dead comment (implemented below)

src/vunnel/providers/nvd/__init__.py
```diff
@@ -38,6 +38,7 @@ def __str__(self) -> str:
 
 
 class Provider(provider.Provider):
+    # TODO: why not int(schema.NVDSchema().major_version) (@weston)
```
wagoodman, line 41: Before we can merge this we should reconcile what version is used (both here and in the listing file)

src/vunnel/cli/config.py
```diff
@@ -26,6 +33,22 @@ class Providers:
     ubuntu: providers.ubuntu.Config = field(default_factory=providers.ubuntu.Config)
     wolfi: providers.wolfi.Config = field(default_factory=providers.wolfi.Config)
 
+    # TODO: try to move this to the toplevel config and make this an init-only var
```
wagoodman, line 36: I'll try to get this migrated, but if I'm out and don't get to it at least this should be moved to a sub-section here called `common` (or a better name) so future common options aren't siblings of other provider configurations.

src/vunnel/providers/nvd/__init__.py
wagoodman: this has been changed to conceptually split up `version` of a provider (which includes processing behavior + results data shape) and the `distribution_version` of a provider (which strictly includes results data shape).

tests/unit/cli/test_cli.py
```diff
@@ -125,6 +128,9 @@ def test_config(monkeypatch) -> None:
     runtime:
       existing_input: keep
       existing_results: delete-before-write
+      import_results_enabled: null
```
wagoodman, line 131: shouldn't this be `False`?

wagoodman (approved):

tests/unit/cli/test_cli.py
willmurphyscode: This was addressed in a later push.

wagoodman (approved):

PR #527, opened 2024-03-29
fix: lower retry threshold and continue on 403 for amazon linux feed

spiffcs:
## Summary 
The amazon linux rss feeds for vulnerabilities currently has multiple 403'ed entries for the following feeds:
- https://alas.aws.amazon.com/AL2/alas.rss
- https://alas.aws.amazon.com/AL2022/alas.rss
- https://alas.aws.amazon.com/AL2023/alas.rss

This PR is a draft hack that get's the amazon linux provider to completion while warning on links that are unable to be reached.

For more information about this failure see:
https://github.com/anchore/vunnel/actions/runs/8477396093

Note: for that run a single entry failed causing the whole process to exit. On this branch, users will see multiple failures of entries, but given the new exception clause the provider will `continue` to try other entries in the rss feed.

I do not think we should merge or consider this PR until we get an answer from aws security on the current state of the feeds and if any updates are expected surrounding the 403'ed records.

It's also worth noting that given the current state of the Amazon Linux Feeds we're missing an entire namespace:
```
{'amazon:distro:amazonlinux:2022'}
```

This is missing because every entry in the following feed is 403'ed:
https://alas.aws.amazon.com/AL2022/alas.rss

### DIFF
--- a/src/vunnel/providers/amazon/parser.py
+++ b/src/vunnel/providers/amazon/parser.py
@@ -92,13 +92,13 @@ class Parser:
                 content = fp.read()
             return content  # noqa: RET504
         try:
-            r = http.get(alas_url, self.logger, timeout=self.download_timeout)
+            r = http.get(alas_url, self.logger, retries=0, timeout=self.download_timeout)
             content = r.text
             with open(alas_file, "w", encoding="utf-8") as fp:
                 fp.write(content)
             return content
         except Exception:
-            self.logger.exception(f"error downloading data from {alas_url}")
+            self.logger.warning(f"error downloading data from {alas_url}")
             raise
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


willmurphyscode:
We need to be careful that `vunnel` can't exit zero and emit a completely broken result set.

Today, vunnel exits non-zero when it fails to build a new result set for a provider, which causes grype-db to re-use the previous day's cached results. After this change, if due to a misconfiguration or something, a large percentage of the listings raise 403, vunnel could write a result set with big gaps in it but still exit 0, causing the result set with big gaps to be used in grype-db, instead of the result set from the previous day.

src/vunnel/providers/amazon/parser.py
```diff
@@ -92,13 +92,13 @@ def _get_alas_html(self, alas_url, alas_file, skip_if_exists=True):
                 content = fp.read()
             return content  # noqa: RET504
         try:
-            r = http.get(alas_url, self.logger, timeout=self.download_timeout)
+            r = http.get(alas_url, self.logger, retries=0, timeout=self.download_timeout)
```
willmurphyscode, line 95: I think `0` is too few retries.

This also doesn't handle 403s specifically, which I thought was the intent of the PR. To handle specific status codes differently from other, you can pass the argument `status_handler` to http.get.

PR #534, opened 2024-04-03
fix: os.rename -> shutil.move

westonsteimel:
Replace instances of `os.rename` with `shutil.move` to handle the possibility of the destination being a different filesystem.  If the source and destination are on the same filesystem, `os.rename` is used per https://docs.python.org/3.11/library/shutil.html#shutil.move

### DIFF
--- a/src/vunnel/workspace.py
+++ b/src/vunnel/workspace.py
@@ -149,11 +149,12 @@ class Workspace:
         utils.silent_remove(os.path.join(self.path, METADATA_FILENAME))
         utils.silent_remove(os.path.join(self.path, CHECKSUM_LISTING_FILENAME))
 
-    def clear_results(self) -> None:
+    def clear_results(self, recreate_results_dir: bool = True) -> None:
         if os.path.exists(self.results_path):
             self.logger.debug("clearing existing results")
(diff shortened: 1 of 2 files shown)


src/vunnel/workspace.py
```diff
@@ -243,17 +244,17 @@ def overlay_existing(self, source: str, move: bool = False) -> None:
                 os.makedirs(os.path.dirname(dst), exist_ok=True)
 
                 if move:
-                    os.rename(src, dst)
+                    shutil.move(src, dst)
```
willmurphyscode, line 247: Actually, we might need the same change here, and it's just not failing a test.

westonsteimel: Is `overlay_existing` actually used anywhere?  I can't find any invocations of it.  Maybe we can just remove it?

willmurphyscode: I don't see it used anywhere. I think it was added by mistake in https://github.com/anchore/vunnel/pull/520, since that PR went through a lot of iterations.

I've pushed a commit removing the function.

spiffcs (approved):

PR #578, opened 2024-05-20
fix: update vunnel provider date and option to skip listing check

asomya:
this commit updates the provider date when vunnel is importing results and an options to skip date checks

### DIFF
--- a/src/vunnel/provider.py
+++ b/src/vunnel/provider.py
@@ -67,6 +67,8 @@ class RuntimeConfig:
     existing_results: ResultStatePolicy = ResultStatePolicy.KEEP
     # the format the results should be written in
     result_store: result.StoreStrategy = result.StoreStrategy.FLAT_FILE
+    # skip checks for newer archive if true (always download latest)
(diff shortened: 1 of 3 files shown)


src/vunnel/provider.py
```diff
@@ -192,8 +194,14 @@ def _fetch_or_use_results_archive(self) -> tuple[list[str], int, datetime.dateti
         if not latest_entry:
             raise RuntimeError("no listing entry found")
 
-        if self._has_newer_archive(latest_entry=latest_entry):
+        if self._has_newer_archive(latest_entry=latest_entry) or self.runtime_cfg.skip_newer_archive_check:
```
wagoodman, line 197:
```suggestion
        if self.runtime_cfg.skip_newer_archive_check or self._has_newer_archive(latest_entry=latest_entry):
```
I think flipping the order has the added benefit of not calling `self._has_newer_archive()` if the config flag is already trie.

wagoodman (approved):

PR #585, opened 2024-05-23
feat: add version range to mariner provider

willmurphyscode:
Previously, this provider could only include upper bounds on ranges.

Note that this will also require a grype-db PR to start putting the new field in the database.


src/vunnel/providers/mariner/model/generated.py
```diff
@@ -270,13 +302,15 @@ class Meta:
         default=None,
         metadata={
             "type": "Attribute",
-        }
+            "required": True,
+        },
     )
-    criterion: Optional[Criterion] = field(
-        default=None,
+    criterion: List[Criterion] = field(
```
willmurphyscode, line 308: This is the important change to the generated model. It might be worth backing out all. the `required: True`, since it's noisy and might lead to a failure to parse the whole OVAL file if there's one malformed record. (The intended behavior of Vunnel is to skip malformed records but carry on and still generate some data.)

My understanding is that right now there will always be exactly 2 items on the list, at least right now.

src/vunnel/providers/mariner/model/generated.py
wagoodman, line 1: there are a lot of `required: True` statements added for unrelated fields? Did something change about the generation process?

wagoodman: nvm -- just saw your comment

wagoodman (approved):

src/vunnel/providers/mariner/model/generated.py
willmurphyscode: This Vunnel parser is written to ignore the failure to parse individual definitions anyway: https://github.com/anchore/vunnel/blob/main/src/vunnel/providers/mariner/parser.py#L45

Based on some experimentation, the new `required: True` annotations are showing up because of bumping the `xsdata` package to version 23.5.

willmurphyscode: Based on offline discussion, we will pin `xsdata` package back to the version that generated these models to avoid introducing unnecessary drift in the models. (Putting `required: True` everywhere has no benefit on a field of type `Optional`, since given a populated model we have to check for `None`, and failing validation for this reason would be unfortunate, since the code defends against these `None`s anyway.)

pyproject.toml
```diff
@@ -54,7 +54,7 @@ orjson = "^3.8.6"
 SQLAlchemy = ">= 1.4.46, < 2.0"  # note: 1.4.x currently required for enterprise
 mergedeep = "^1.3.4"
 importlib-metadata = "^7.0.1"
-xsdata = {extras = ["cli", "lxml", "soap"], version = ">=22.12,<25.0"}
+xsdata = {extras = ["cli", "lxml", "soap"], version = "=22.12"}
```
spiffcs, line 57: Dev note: we might have to unpin this eventually to generate the models at a later date, but this pin keeps the diff less noisy until we want to tackle this issue

PR #588, opened 2024-05-24
fix: record state after timestamp update

asomya:
fixing a bug where state was not recorded after updating timestamp


wagoodman (commented):
I think there is still a need to change `_update` (+ test) to consider not only `count > 0` but also `stale` (probably `stale or count > 0`)

svietry:
Is there a unit test that can be added for this case?
@asomya

wagoodman (approved):

PR #597, opened 2024-06-06
 feat(ubuntu): treat needs-triage as vulnerable until determination made

westonsteimel:


westonsteimel:
This was an ask from Josh and others from the community ~7 months ago and seems reasonable.  I had made a start at it but then got caught up in other things and forgot about it.  It has come up again recently, so attempting again.  I'm not sure if it is as simple as setting `needs-triage` to True here, but hopefully it will be

wagoodman (approved):

PR #601, opened 2024-06-11
feat: add ubuntu 24.10 codename to version mapping

westonsteimel:
The codename for Ubuntu 24.10 is `Oracular Oriole` per https://wiki.ubuntu.com/Releases, so this gets the mapping to version in place well in advance


juanjsebgarcia (approved):
spectacular

PR #622, opened 2024-07-09
Add two CVE IDs to the severity override list

joshbressers:


src/vunnel/providers/debian/parser.py
```diff
@@ -355,9 +355,10 @@ def _normalize_json(self, ns_cve_dsalist=None):  # noqa: PLR0912,PLR0915,C901
 
                             # HACK: when we can represent per-package severity or have a good mechanism
                             # for overriding upstream data, we should take this out.
-                            if vid == "CVE-2023-44487":
+                            severity_override = ["CVE-2023-44487", "CVE-2020-35525", "CVE-2020-35527"]
```
wagoodman, line 358:
```suggestion
                            severity_override = {"CVE-2023-44487", "CVE-2020-35525", "CVE-2020-35527"}
```
this should perform a little better as a set

wagoodman (approved):
added a comment based on efficiency, but otherwise 👍

src/vunnel/providers/debian/parser.py
joshbressers: A set is a good point. I'll add it in my next commit, we need to add another ID before merging

Issue #644, opened 2024-07-30
chore: replace frequenty used dict literals with data classes


willmurphyscode:
For example:

https://github.com/anchore/vunnel/blob/c615dcae89ec4b84d7a51a62d77c69a6a496af9e/src/vunnel/utils/vulnerability.py#L7-L14 gets `copy.deepcopy`-ed around a lot, as a template for a dict literal. The code would be cleaner and have more useful type hinting if this were replaced with a data class.

https://github.com/anchore/vunnel/blob/c615dcae89ec4b84d7a51a62d77c69a6a496af9e/src/vunnel/utils/vulnerability.py#L16-L27 is another example

Examples of copying https://github.com/anchore/vunnel/blob/c615dcae89ec4b84d7a51a62d77c69a6a496af9e/src/vunnel/utils/oval_parser.py#L160-L170

and https://github.com/anchore/vunnel/blob/c615dcae89ec4b84d7a51a62d77c69a6a496af9e/src/vunnel/providers/wolfi/parser.py#L103-L109

willmurphyscode:
We want to make sure that whatever solution we use is not vulnerable to crashing because a field is added, so maybe skipping straight to data classes is the wrong approach. I'll research some additional approaches.

PR #645, opened 2024-07-30
chore: move debian parser from vulnerability dict to dataclass

willmurphyscode:
Previously, the parser was using a deep copy of a dict literal to make each new instance of the vulnerability it was emitting. Vunnel has since added a dataclass to represent vulnerabilities. Switch to using that dataclass.

See #644


willmurphyscode:
It looks like the quality gate is failing because somehow there are a mix of Vulnerability dataclasses and dicts in `vuln_records`:

```
   File "/home/runner/work/vunnel/vunnel/src/vunnel/providers/debian/parser.py", line 553, in get
    vuln_record["Vulnerability"].Severity = "Unknown"
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'dict' object has no attribute 'Severity'
```

westonsteimel:
> It looks like the quality gate is failing because somehow there are a mix of Vulnerability dataclasses and dicts in `vuln_records`:
> 
> ```
>    File "/home/runner/work/vunnel/vunnel/src/vunnel/providers/debian/parser.py", line 553, in get
>     vuln_record["Vulnerability"].Severity = "Unknown"
>      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> AttributeError: 'dict' object has no attribute 'Severity'
> ```

@willmurphyscode , could that be due to the "legacy" record path?

willmurphyscode:
@westonsteimel that seems likely. I knew the Debian provider was downloading cached data. There must be a place where it's parsing that as dict instead of Vulnerability. 

Do you think it's safe to change that path? I know there's some information we care about that's only available in the legacy cache, so I'm wondering if I should abandon this for the Debian provider in particular. But I can't think of a reason that writing those records back to the cache as Vulnerability would matter - the cache is just JSON if I recall correctly.

westonsteimel:
It should be safe to do.  Alex recently changed it so that the legacy cache is a sqlite db with exactly the vunnel result shape

willmurphyscode:
This is waiting on #647 so that any differences in snapshots will be possible to review.

westonsteimel:
Oh, one danger with python dataclass that we ran into with enterprise is that it can't handle new properties being added to the json.  So, for instance, if we added a new property to the os schema (non breaking change) and published that and then consumed in proxy mode, old vunnel clients using dataclass would actually fail due to an additional unknown property being present

willmurphyscode:
@westonsteimel does that mean we should hold off on this type of change until the next release?

willmurphyscode:
There's doubt about whether this is the right direction and it's not on top of the priority pile. We can revisit this if it comes up again.

PR #650, opened 2024-08-02
fix: grype_db #344 missing CVE/Package associations

BenoitGui:
Fix https://github.com/anchore/grype-db/issues/344, some package where missing from sles database because of an error in the parser.
The code avoid to crush old packages list when we have multiple criteria for the same os version. 
Thanks to @spiffcs and @willmurphyscode for the help


willmurphyscode:
Thanks for the fix @BenoitGui! I'm going to try to add a unit test or two so that we don't have a regression here, and then we should be able to get this in.

westonsteimel (approved):

PR #672, opened 2024-08-30
chore: switch to new yardstick validate

willmurphyscode:
Pull in changes from anchore/yardstick#380.

## TODOs

- [x] actually delete `tests/quality/gate.py`, since it is no longer invoked
- [x] update docs to remove references to `test/quality/gate.py` and replace with yardstick
- [x] make sure the use of different result sets for each provider hasn't messed up some CI caching like `make sboms` or something.
- [x] pin to release of yardstick, not a commit sha.


wagoodman (approved):

tests/quality/configure.py
```diff
@@ -601,8 +615,8 @@ def build_db(cfg: Config):
         subprocess.run(["vunnel", "-v", "run", provider], check=True)
 
     logging.info("building DB")
-    subprocess.run([GRYPE_DB, "build", "-v"], check=True)
-    subprocess.run([GRYPE_DB, "package", "-v"], check=True)
+    subprocess.run([GRYPE_DB, "build", "-v", "-c", ".grype-db.yaml"], check=True)
```
willmurphyscode, line 618: @wagoodman this is part of the fix that I wanted your feedback on. My understanding of how `grype-db` searches for config files says I shouldn't have to pass that, but it fails unless I do (It seems to find the `.grype-db.yaml` that gets written at the repo root rather than the one that gets written in `tests/quality`.

PR #730, opened 2024-11-14
Add processor to workspace state

wagoodman:
This adds vunnel version information to workspace output, so we know which version of vunnel did the processing of the upstream data. This is related to [grype DB v6 work](https://github.com/orgs/anchore/projects/22/views/23?pane=issue&itemId=80129876&issue=anchore%7Cgrype%7C2125) so that we have more detail on a provider-level on how the data was handled upstream.

### DIFF
--- a/schema/provider-workspace-state/schema-1.0.3.json
+++ b/schema/provider-workspace-state/schema-1.0.3.json
@@ -1,0 +1,84 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "title": "provider-workspace-state",
+  "description": "describes the filesystem state of a provider workspace directory",
+  "properties": {
+    "provider": {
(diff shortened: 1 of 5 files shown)


schema/provider-workspace-state/schema-1.0.3.json
```diff
@@ -0,0 +1,84 @@
...
+      "description": "version describing the result data shape + the provider processing behavior semantics"
+    },
+    "distribution_version": {
+      "type": "integer",
+      "description": "version describing purely the result data shape"
+    },
+    "processor": {
+      "type": "string",
+      "description": "the application that created the workspace, formatted as 'name@version'"
+    },
```
wagoodman, line 54: This is what was added relative to the last schema

kzantow (approved):

PR #738, opened 2024-11-26
Wire up retry count config to NVD provider

wagoodman:
This PR primarily wires up a new `request_retry_count` option to NVD with a default of 10 (a rather high number since the API tends to show issues). This also adjusts the jitter and display behavior of the `http.get` helper:
- now the sleep interval (and not the backoff) is shown when in the "attempting in X seconds" log message
- jitter applies to the first sleep interval (not the second)


kzantow (approved):

wagoodman:
Bypassed checks since NVD API is in a degraded state

PR #747, opened 2024-12-19
feat(github): pull erlang advisory data

westonsteimel:


westonsteimel:
Nevermind, this one needs more thought as to how to populate both the `erlang` and `elixir` namespaces for grype db v5.  It may be that the responsibility of copying the data to another namespace should be in grype-db, in which case this simple vunnel change will be fine

willmurphyscode:
@westonsteimel should we just make erlang / elixir v6 only? That's what we're doing to simplify adding other new data.

PR #758, opened 2025-01-09
feat: RHEL CSAF provider

### LINKED ISSUE #323: Switch RedHat vulnerability provider from OVAL to CSAF

The RedHat provider currently uses the v2 OVAL files for RedHat vulnerability data; however, those will only continue to be updated until the end of 2024.  We need to transition to using the new CSAF endpoints per https://www.redhat.com/en/blog/future-red-hat-security-data

willmurphyscode:
Add a new provider, rhel_csaf, that downloads rhel vulnerability data from the CSAF VEX JSON that RedHat publishes. Move the existing rhel provider to rhel_legacy. Add a top-level rhel provider that can be configured to delegate to rhel_legacy or rhel_csaf. Add some dataclasses to deserialize CSAF JSON into.

Fixes #323

- [ ] - pass down configs and logging
- [x] - unit tests for uber provider
- [ ] - unit tests for csaf provider
- [ ] - unit tests for transformer
- [ ] - test and handle not-fixed vs won't-fix
- [x] - tweak quality gate
- [ ] - README explaining how new provider works
- [ ] - chase down WARN unable to merge matches ... from grype
- [x] - figure out about incremental updates and `skip_if_exists`
- [ ] - handle cvss v2 scores
- [ ] - better way to tweak quality gate


willmurphyscode:
Looks like this failed with no space left on device: https://github.com/anchore/vunnel/actions/runs/12682455877/job/35347881901?pr=758#step:4:1848

wagoodman:
re: no space, you can use `runs-on: ubuntu-22.04-4core-16gb` both

https://github.com/anchore/vunnel/blob/697e8a2afc748e65a3852566c4e41ff5588427b2/.github/workflows/pr-quality-gate.yaml#L42

and

https://github.com/anchore/vunnel/blob/697e8a2afc748e65a3852566c4e41ff5588427b2/.github/workflows/nightly-quality-gate.yaml#L36

src/vunnel/providers/rhel_csaf/transformer.py
```diff
@@ -0,0 +1,245 @@
...
+                            # pick those when we do the component packages.
+                            continue
+                        module = purl.namespace.removeprefix("redhat/")
+                else:  # "rpm"
+                    module_pid = None
+                    parent = csaf.product_tree.parent(qpi)
+                    if not parent:
+                        # TODO: log a warning - got a leaf product with no parent distro or module
+                        continue
+                    grand_parent = csaf.product_tree.parent(parent or "NONE")
```
willmurphyscode, line 182: This is probably wrong. Why not `None`?

tests/quality/config.yaml
```diff
@@ -223,7 +223,9 @@ tests:
       - docker.io/anchore/test_images:appstreams-rhel-8-1a287dd@sha256:524ff8a75f21fd886ec7ed82387766df386671e8b77e898d05786118d5b7880b
       - docker.io/anchore/test_images:vulnerabilities-centos@sha256:746d31247006cc06434ce91ccf3523b2c230ff6c378ffed7ca1c60bbb48ea86f
     validations:
-      - *default-validations
+      - <<: *default-validations
+        max_f1_regression: 0.09
```
willmurphyscode, line 227: We really don't want to do this. We need to investigate the particular false positive in the quality gate:

```
   grype[custom-db]@v0.86.1-14-g67ac96f ONLY  mariadb-common@3:10.3.28-1.module_el8.3.0+757+d382997d      CVE-2021-27928  FalsePositive  (this is a new FP 😱)
   grype[custom-db]@v0.86.1-14-g67ac96f ONLY  mariadb@3:10.3.28-1.module_el8.3.0+757+d382997d             CVE-2021-27928  FalsePositive  (this is a new FP 😱)
```

I think we should investigate fixing grype's version comparison so that RPM version `3:10.3.28-1.module_el8.3.0+757+d382997d` compares as equal to `3:10.3.28-1.module+el8.3.0+10472+7adc332a` rather than less. Then neither reference nor candidate will find this CVE.

Specifically https://github.com/anchore/grype/blob/ed98490e966cf215eb785395fe80b79f99e420fa/grype/version/rpm_version.go#L99-L105

This needs to be smarter to understand that centOS and RedHat build numbers are not comparable, so `1.module+el8.3.0` and not `1.module+el8.3.0+10472+7adc332a` is the string being compared. Probably a new `compareRelease` function instead of calling `compareVersion` again.

Issue #742, opened 2024-12-04
Track date timestamps in vulnerability data


wagoodman:
**What would you like to be added**:
Where possible, track the following timestamps for each provider:
- published date
- modified date
- withdrawn date

**Why is this needed**:
This is very helpful in terms of understanding what data has changed in the grype DB, especially when grype DB v6 schema lands.

**Additional context**:

Here's what I see today in terms of providers need work (github and nvd are not listed since they already capture this information and output it as results):

| Provider      | Published | Modified | Withdrawn | Comments                                                                 |
|---------------|-------------------|------------------|-------------------|-------------------------------------------------------------------------|
| **alpine**    | ⚠️                | ⚠️               | ✅                | Use NVD date info (publish + modified + withdrawn) + `aports` git timestamps (publish + modified) |
| **wolfi**     | ✅                | ✅               | ✅                | Port to [using advisories](https://github.com/wolfi-dev/advisories) over secdb   |
| **chainguard**| ✅                | ✅               | ✅                | Port to [using advisories](https://edu.chainguard.dev/chainguard/chainguard-images/working-with-images/security-advisories/how-to-use/) over secdb   |
| **amazon**    | ✅                | ✅               | ❌                | Use XML `pubdate` + HTML span info      |
| [**debian**](#732)    | ✅                | ❌               | ❌                | Dates listed for each DSA . **Important:** Legacy distros not covered |
| **mariner**   | ✅                | ❌               | ❌                | Use `advisory_date` field in OVAL XML       |
| **oracle**    | ✅                | ❌               | ❌                | Use `issued` field in `elsa-all` XML     |
| **rhel**      | ✅                | ✅               | ❌                | Use `issued` and `updated` in OVAL XML   |
| **sles**      | ✅                | ✅               | ❌                | Use `issued` and `updated` in OVAL XML   |
| **ubuntu**    | ⚠️        | ⚠️               | ❌                | Use git timestamp for each commit when processing (modified). Published requires more effort |

✅ - data is trivially accessible from existing input data already being downloaded
⚠️ - possible to associate the data but will require a little more work (or downloading more data sources)
❌ - not clear where to get this information from

zhill:
I think there is a good use-case for specifically tracking when fix information is added to a record outside of just the broader "Modified" category if that information available.

wagoodman:
Good call -- the v6 schema has a spot for this but it is not really leveraged yet, I'll update this table...

PR #760, opened 2025-01-10
Add date information to OS schema

wagoodman:
Related to #742, adds the following fields to the os schema:
- Metadata.Updated: date the vuln record was updated
- Metadata.Withdrawn: date the vuln record was withdrawn
- FixedIn.Issued: date the fix was made available

Note that `Metadata.Issued` (which is when the vulnerability was first published) already exists in the schema.

### DIFF
--- a/tests/unit/providers/alpine/test-fixtures/snapshots/3.15/cve-2018-1071.json
+++ b/tests/unit/providers/alpine/test-fixtures/snapshots/3.15/cve-2018-1071.json
@@ -19,5 +19,5 @@
       "Severity": "Unknown"
     }
   },
-  "schema": "https://raw.githubusercontent.com/anchore/vunnel/main/schema/vulnerability/os/schema-1.0.0.json"
+  "schema": "https://raw.githubusercontent.com/anchore/vunnel/main/schema/vulnerability/os/schema-1.0.2.json"
 }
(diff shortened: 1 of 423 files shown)


wagoodman:
This is the actual change
```diff
# $ diff schema/vulnerability/os/schema-1.0.1.json schema/vulnerability/os/schema-1.0.2.json
98a99,102
>                 },
>                 "Issued": {
>                   "type": "string",
>                   "description": "date the fix was made available"
117c121,122
<               "type": "string"
---
>               "type": "string",
>               "description": "date the vulnerability was published"
118a124,131
>             "Updated": {
>               "type": "string",
>               "description": "date the vulnerability was last updated"
>             },
>             "Withdrawn": {
>               "type": "string",
>               "description": "date the vulnerability was withdrawn"
>             },
```

kzantow (approved):
There are a lot of changes here! You definitely know better than me what's going on, so I'll just leave a couple observations:
* should this have a `schema.json`, like we do in syft which is always the current and would have ended up showing a diff here? (and could be referred to by commit; unsure if that makes anything really easier)
* it would be ideal if bumping the schema version wouldn't require updating all the test fixtures like this -- maybe some process somewhere updates the values while reading them in test fixtures or migrates the files somehow if it was an easy process

wagoodman:
Yeah, I agree with both observations -- that's going to require more refactoring of test utils and cases + adding some automation around how the schema is bumped (which does not exist). This PR isn't the right time to introduce those things (but am all for it).

PR #763, opened 2025-01-13
Add explicit version information to emitted records

wagoodman:
Today we have the namespace field which conveys "ID:VERSION" for distro information. This field is a hold over from how anchore enterprise used to function. The downside is that the current version resolution for some operating systems does not consider the minor or patch version, which is conveyed in some OS fix information (such as redhat and oracle linux). This PR nudges the existing OS schema forward by adding a specific OS field with OS ID (aligning with /etc/os-release ID) and Version properties (which should be as specific as possible relative to the fix information). All providers have been updated to use this new field (ideally we'll remove the Namespace fields as a breaking change in the future).

I've additionally added architecture qualifiers to the fix information as well, but have not updated any provider to use this new field.


schema/vulnerability/os/schema-1.0.3.json
wagoodman, line 1:
```diff
diff schema/vulnerability/os/schema-1.0.2.json schema/vulnerability/os/schema-1.0.3.json
72a73,89
>                 "OS": {
>                   "type": "object",
>                   "properties": {
>                     "ID": {
>                       "type": "string"
>                     },
>                     "Version": {
>                       "type": "string"
>                     }
>                   }
>                 },
>                 "Architectures": {
>                   "type": "array",
>                   "items": {
>                     "type": "string"
>                   }
>                 },
107d123
<                 "NamespaceName",
191d206
<         "NamespaceName",

```

westonsteimel:
Hmm, what about previous minor versions of the same release line though?  For instance, if there is a fix available in the rhel 8.7 appstream but the prior 8.x appstreams should still be considered vulnerable because there won't be a fix released to them?

PR #765, opened 2025-01-17
chore: switch to `ruff format` over `black`

willmurphyscode:
We've had both `black` and `ruff` making changes to the code on lint/pre-commit for a long time. 

Ruff now [claim that their formatter is a drop-in replacement for `black`](https://docs.astral.sh/ruff/faq/#how-does-ruffs-formatter-compare-to-black).

This PR switches `make format` from using `black` to using `ruff format`. 

The rest of the diff is the result of running `make format` on the code after the switch, which look like only line length changes.

When we had `black` and `ruff`, both had a line-length setting, and black was set to 130 but to ignore some files, and ruff was set to 150 but to ignore no files. In `ruff format`, there's no way to set the line length to be different for linting and formatting, so some lines are changing from 130 to 150. So there used to be some files with a 130 line length limit and some files with a 150 line length limit. Now all files are at 150.

We could also standardize on 130. I'll defer to whatever the team wants here.

Also, `black` used to allow (require?) a blank line between `class Foo:` and `__some_var__ = whatever`. `ruff format` by default takes this line out. I haven't figured out a way to tell `ruff format` to allow this, so these blank lines are going away.

The total speed up benefit of this is not as much as I'd hoped - `make static-analysis` is pretty variable in how long it takes, but it seems between 200 and 500 millis faster on this branch, out of a total of about 4000 millis, so about maybe 10% faster). Still, this gets run a lot, so a small speed-up is welcome.

I think the real benefit is having only one tool that care about line length. Having a list of strings as exclude paths instead of a multi line regex is nice too.


.pre-commit-config.yaml
```diff
@@ -57,8 +57,8 @@ repos:
     hooks:
 
       # note: this is used in lieu of autopep8 and yapf
-      - id: black
-        name: black
+      - id: format
```
spiffcs, line 60: Good change to generalize this rather than use the tool name

pyproject.toml
```diff
@@ -77,7 +76,7 @@ types-requests = "^2.28.11.7"
 mypy = "^1.1"
 radon = ">=5.1,<7.0"
 dunamai = "^1.15.0"
-ruff = ">=0.5.1,<0.9.2"
+ruff = "^0.9.2"
```
spiffcs, line 79: confirmed v0.9.2 is latest from 5 days ago

src/vunnel/providers/nvd/api.py
```diff
@@ -39,12 +39,8 @@ def __init__(
     def cve_history(
         self,
         cve_id: str | None = None,
-        results_per_page: (
-            int | None
-        ) = None,  # from api docs: "it is recommended that users of the CVE API use the default resultsPerPage value"
-        change_start_date: (
-            str | datetime.datetime | None
-        ) = None,  # note: if you specify a changeStartDate, you must also specify a changeEndDate
+        results_per_page: (int | None) = None,  # from api docs: "it is recommended that users of the CVE API use the default resultsPerPage value"
```
spiffcs, line 42: This looks much better

spiffcs (approved):

PR #784, opened 2025-03-06
Port from poetry to uv

wagoodman:
Additionally migrates from Makefile to Taskfile/Binny.


Makefile
```diff
@@ -1,193 +1,43 @@
...
-	mkdir -p $(BIN_DIR)
-
-## Development targets #################################
-
-.PHONY: dev
-dev:  ## Get a development shell with locally editable grype, grype-db, and vunnel repos
-	@DEV_VUNNEL_BIN_DIR=$(ABS_BIN_DIR) .github/scripts/dev-shell.sh $(provider) $(providers)
-
-.PHONY: build-grype
-build-grype: $(BIN_DIR) ## Build grype for local development
```
willmurphyscode, line 84: It looks like `build-grype` and `build-grype-db` didn't make it to the task file. Is this intentional?

Makefile
```diff
@@ -1,193 +1,43 @@
...
-	@cd $(GRYPE_PATH) && go build -o $(ABS_BIN_DIR)/grype ./cmd/grype
-
-.PHONY: build-grype-db
-build-grype-db: $(BIN_DIR) ## Build grype-db for local development
-	@cd $(GRYPE_DB_PATH) && go build -o $(ABS_BIN_DIR)/grype-db ./cmd/grype-db
-
-.PHONY: update-db
-update-db: check-dev-shell ## Build and import a grype database based off of the current configuration
```
willmurphyscode, line 92: It looks like `update-db` also didn't make it to the task file. Is this intentional?

Makefile
wagoodman: not intentional, I'll incorporate

Makefile
wagoodman: I missed all of the dev targets, I'll push an update

willmurphyscode (approved):

PR #794, opened 2025-03-14
feat: rocky linux OSV provider

willmurphyscode:
Add Rocky Linux OSV provider.

This provider will use the grype-db OSV transformer in progress at anchore/grype-db#217.

This provider uses the Rocky Linux Apollo API, documented at https://distro-tools.rocky.page/apollo/openapi/#get-apiv3osv

This adds OSV schema 1.3.1 because that is the schema found in that data.

How to integrate this data with RedHat data to supplement matching on Rocky images is something we will have to figure out, but likely outside the scope of this PR.


src/vunnel/providers/rocky/__init__.py
```diff
@@ -0,0 +1,73 @@
...
+from vunnel import provider, result, schema
+
+from .parser import Parser
+
+if TYPE_CHECKING:
+    import datetime
+
+
+@dataclass
+class Config:
```
wagoodman, line 15: this needs to be wired up to the application config class + tests added for that change

src/vunnel/providers/rocky/client.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def __init__(
+        self,
+        download_path: str,
+        logger: logging.Logger | None = None,
+        rocky_versions: list[str] | None = None,
+        api_host: str = "https://apollo.build.resf.org",
+    ):
+        if rocky_versions is None:
+            rocky_versions = ["8", "9"]
```
wagoodman, line 22: is there a way to make this more dynamic? Some providers hard code this kind information either because it's not possible to dynamically figure it or (most of the time) we didn't want to have enterprise deployments start failing when unexpected data started flowing from a new version. The latter point is no longer a risk. 

Consider this comment not a request to change anything now but more on how we can improve these kinds of case systematically in the future (may be worth an issue too).

willmurphyscode: It turns out this wasn't doing anything, so I just removed it.

PR #814, opened 2025-04-30
Add support for MinimOS

Daniel-Wachter:
Hi,
I’m part of the [Minimus](https://minimus.io/) team. Minimus delivers secure, minimal container images with auto-generated SBOMs and real-time vulnerability threat intelligence to help reduce vulnerability risk.
We also maintain a minimal operating system called MinimOS. We’ve been publishing our security advisories in a secdb feed and would like to contribute it as a new security data source.

Details:
The feed URL: https://packages.mini.dev/advisories/secdb/security.json

The format closely mirrors Alpine's secdb, but it's unversioned—similar to Alpine's edge feed.

Discussion ref: https://github.com/aquasecurity/trivy/discussions/8666

Thanks for your consideration!


willmurphyscode:
Hi @Daniel-Wachter thanks for the contribution! It looks good. I will add some feedback (or, more likely, push some small additional type hinting) to this PR and start looking at your other PRs as well.

However, for end to end testing it would be very helpful to have access to a minimOS OCI image that we could scan and ensure it's valid. Is it possible for you to permit unauthenticated pulls of a small image, maybe from an old tag or something? 

The real test of this change is to build a grype database and a copy of grype incorporating your changes, point it at a real OCI image that is expected to have a couple vulnerabilities (so an older tag/sha would be great), and make sure grype finds them. This tests the whole stack - finding and identifying packages and their versions, mapping the distro to this vulnerability provider, etc, and it is difficult to be confident of adding a new vulnerability provider without it.

orzlc:
@willmurphyscode thanks, we pushed two images to an unauthenticated DockerHub account, and we plan to keep them live. The two images have different versions of the same packages (the second is newer). Both of them have vulnerabilities, the first one has more.

First digest: dimastopelmini/forgrype@sha256:ebe0c6ca122deef072c29be2f915130e5c8b4c277ad5ef551385f
Second digest: dimastopelmini/forgrype@sha256:653c8980c63a9ac403a3b9f56a08f43f929432ece69894423c165b4d61d3dcdb

PR #818, opened 2025-05-24
add eol

tomersein:


tomersein:
@wagoodman any updates?

westonsteimel:
Apologies @tomersein , I'm still buried under some other data work so will still be a bit before I get to this.  I did notice that there isn't yet a json schema file defined under `/schema/` for this though

PR #843, opened 2025-08-11
feat(debian): add future mappings for debian 14 and 15

westonsteimel:
Adds the expected mappings for dbeian 14 and 15 now that 13 is released


popey (approved):
LGTM

PR #852, opened 2025-08-21
Add fix date processing for github

wagoodman:
This adds adding fix-date information based on first-observed fix information to records for the github provider.

Partially implements #742

PR #853, opened 2025-08-21
Add fix date processing for alpine

wagoodman:
This adds adding fix-date information based on first-observed fix information to records for the alpine provider.

Partially implements #742 

## PR Stack
- #852

### DIFF
--- a/src/vunnel/providers/alpine/parser.py
+++ b/src/vunnel/providers/alpine/parser.py
@@ -17,6 +17,7 @@ if TYPE_CHECKING:
     import requests
 
     from vunnel import workspace
+    from vunnel.tool import fixdate
 
 namespace = "alpine"
 feedtype = "vulnerabilities"
@@ -53,14 +54,16 @@ class Parser:
     _link_finder_regex_ = re.compile(r'href\s*=\s*"([^\.+].*)"')
     _security_reference_url_ = "https://security.alpinelinux.org/vuln"
 
-    def __init__(
+    def __init__(  # noqa: PLR0913
         self,
         workspace: workspace.Workspace,
+        fixdater: fixdate.Finder | None = None,
         logger: logging.Logger | None = None,
         download_timeout: int = 125,
         url: str | None = None,
         security_reference_url: str | None = None,
     ):
+        self.fixdater = fixdater
         self.download_timeout = download_timeout
         self.source_dir_path = os.path.join(
             workspace.input_path,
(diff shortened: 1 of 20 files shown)


src/vunnel/providers/alpine/parser.py
```diff
@@ -249,10 +255,27 @@ def _normalize(self, release, dbtype_data_dict):  # noqa: C901, PLR0912
 
                             # SET UP fixedins
                             fixed_el = {}
+                            ecosystem = f"{namespace}:{release}"
```
willmurphyscode, line 258: @wagoodman a small difference: lots of similar parsers explicitly have `str(release)` in the related PRs. Just wanted to make sure that isn't needed here. (I don't have any reason to suspect it is besides the difference between the other similar PRs).

wagoodman: the goal is to have the namespaceName as-is today, which matches the NAME:VERSION pattern.... in the diff below it started as `fixed_el["NamespaceName"] = namespace + ":" + str(release)` and changed to `ecosystem`, and ecosystem is `f"{namespace}:{release}"` -- so no change. Every provider was written differently, so some have a release variable in place that is the namespace name (confusing!)

wagoodman: also, in terms of what the data should expect when querying for a date:
```
+---------------+----------+--------------+----------+------------+--------------+---------------------+------------+----------+--------+-------------+
| vuln_id       | provider | package_name | full_cpe | ecosystem  | fix_version  | first_observed_date | resolution | source   | run_id | database_id |
+---------------+----------+--------------+----------+------------+--------------+---------------------+------------+----------+--------+-------------+
| CVE-2016-9401 | alpine   | bash         |          | alpine:3.2 | 4.3.33-r2    | 2020-08-07          | day        | grype-db | 1      | 1           |
| CVE-2017-3136 | alpine   | bind         |          | alpine:3.2 | 9.10.4_p8    | 2020-08-07          | day        | grype-db | 1      | 1           |
| CVE-2017-3137 | alpine   | bind         |          | alpine:3.2 | 9.10.4_p8    | 2020-08-07          | day        | grype-db | 1      | 1           |
| CVE-2017-3138 | alpine   | bind         |          | alpine:3.2 | 9.10.4_p8    | 2020-08-07          | day        | grype-db | 1      | 1           |
```

PR #858, opened 2025-08-21
[WIP] Add fix date processing for rocky

wagoodman:
This adds adding fix-date information based on first-observed fix information to records for the rocky provider.

Partially implements #742 

## TODO
- this is NOT meant to be merged since the rocky provider is not used yet

## PR Stack
- #852


tests/unit/utils/test_osv.py
```diff
@@ -0,0 +1,415 @@
...
+            vuln_id="CVE-2023-1234",
+            cpe_or_package="example-pkg",
+            fix_version="2.1.0",
+            ecosystem="npm"
+        )
+
+        # Verify the fix data was added
+        expected_fix = {
+            "version": "2.1.0",
+            "date": "2023-06-15T10:30:45",
```
willmurphyscode, line 203: The test fixtures elsewhere have dates like `"2024-01-14"`. I'm just making sure nothing is expected to truncate this to only the date portion. Just following up on being surprised to see different date format here than in `tests/unit/providers/bitnami/test-fixtures/snapshots/bit-consul-2021-37219.json` above, for example.

wagoodman: nice catch -- updated to just dates

PR #861, opened 2025-08-26
Add fix date processing for NVD

wagoodman:
This adds adding fix-date information based on first-observed fix information to records for the nvd provider.

Partially implements #742 

## PR Stack
- #852


src/vunnel/providers/nvd/manager.py
```diff
@@ -55,36 +58,44 @@ def get(
         last_updated: datetime.datetime | None,
         skip_if_exists: bool = False,
     ) -> Generator[tuple[str, dict[str, Any]], Any, None]:
-        self.overrides.download()
+        """main method to fetch and process NVD vulnerability data.
```
willmurphyscode, line 61: Thanks for all the really clear doc comments in here. It made reviewing this class much easier.

wagoodman: it tripped me up, so figured it needed a sprinkling of comments ✏️

willmurphyscode (approved):

PR #864, opened 2025-08-29
Add fix date processing for redhat

wagoodman:
This adds adding fix-date information based on first-observed fix information to records for the rhel provider.

Partially implements #742

### DIFF
--- a/src/vunnel/providers/rhel/parser.py
+++ b/src/vunnel/providers/rhel/parser.py
@@ -24,6 +24,8 @@ from vunnel.utils.vulnerability import vulnerability_element
 if TYPE_CHECKING:
     import requests
 
+    from vunnel.tool import fixdate
+
     from .rhsa_provider import RHSAProvider
 
 namespace = "rhel"
@@ -54,6 +56,7 @@ class Parser:
     def __init__(  # noqa: PLR0913
         self,
         workspace,
+        fixdater: fixdate.Finder | None = None,
         download_timeout=None,
         max_workers=None,
         full_sync_interval=None,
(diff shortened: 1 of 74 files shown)


tests/unit/providers/rhel/test-fixtures/snapshots/rhel:5/cve-2017-3526.json
```diff
@@ -45,7 +53,11 @@
             "NoAdvisory": true
           },
           "Version": "None",
```
willmurphyscode, line 55: Does this fixture intentionally have a fix available date but no fixed version?

wagoodman: nope that is an oversight

src/vunnel/providers/rhel/parser.py
```diff
@@ -229,6 +233,12 @@ def _sync_cves(self, skip_if_exists=False, do_full_sync=True):  # noqa: PLR0915,
         utils.silent_remove(os.path.join(self.cve_dir_path, self.__last_synced_filename__))
         utils.silent_remove(os.path.join(self.cve_dir_path, self.__cve_download_error_filename__))
 
+        if self.fixdater:
+            if self.skip_download:
+                self.logger.warning("skip download requested, but fix date finder does not support skipping download")
```
willmurphyscode, line 238: This would be nice to add some day, mostly so that when I'm iterating on a provider I don't need to download the fix date info every time. Definitely out of scope for this PR though :)

willmurphyscode (approved):

PR #892, opened 2025-09-22
fix: workspace result_count fails with multiple .db files

jamestexas:
# Fix workspace result_count failure with multiple .db files

## Summary

Fixes 'no such table: results' error in `vunnel status` command when fix-dates feature creates `observed-fix-dates.db` alongside `results.db`.

## Motivation

The `workspace.result_count()` method was querying all `.db` files for a `results` table, but `observed-fix-dates.db` only has a `fixdates` table, causing the query to fail with "no such table: results".

This regression broke the `vunnel status` command for providers using the fix-dates feature, preventing users from checking provider state.

## Changes

- Modified `workspace.result_count()` to only query files ending with `results.db` for vulnerability counts
- Other `.db` files are now treated as regular files (+1 to count) instead of being queried for results
- Added comprehensive regression test reproducing the multiple .db file scenario

## Root Cause

The fix-dates feature (introduced in v0.39.0) creates auxiliary databases in the results directory:

```
results/
├── results.db          ← Main vulnerability data (has "results" table)
├── observed-fix-dates.db ← Fix date tracking (has "fixdates" table)
```

The old logic tried to query **all** `.db` files for vulnerability results, causing crashes when auxiliary databases had different schemas.
**NOTE** If this is desired behavior I am happy to iterate through the db files and combine them if a results table is known, just let me know.

## Testing

- Added `test_result_count_with_multiple_db_files_regression()` that reproduces the exact failure scenario
- Test creates both `results.db` and `observed-fix-dates.db` with different table schemas
- Verifies that only `results.db` is queried for vulnerability counts
- Confirms that auxiliary `.db` files are properly handled without crashes

## Backward Compatibility

This change is fully backward compatible:
- Existing behavior unchanged when only `results.db` exists
- Fix-dates auxiliary databases are properly handled without breaking status command
- No changes to output format or API

Fixes regression introduced in v0.39.0 when fix-dates feature was added.

### DIFF
--- a/src/vunnel/workspace.py
+++ b/src/vunnel/workspace.py
@@ -111,8 +111,14 @@ class State(DataClassDictMixin):
                     if filepath.endswith("results.db"):
                         # open up the sqlite db and count the records in the "results" table
                         db_path = os.path.join(root, filepath)
-                        with sqlite3.connect(db_path) as db:
-                            count += db.execute("SELECT COUNT(*) FROM results").fetchone()[0]
+                        if os.path.exists(db_path):
+                            with sqlite3.connect(db_path) as db:
+                                query = "SELECT COUNT(*) FROM results"
+                                result = db.execute(query).fetchone()
+                                count += result[0]
+                        # if file doesn't exist, treat as regular file
+                        else:
+                            count += 1
                     else:
                         count += 1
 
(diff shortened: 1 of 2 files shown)


westonsteimel:
Thanks for the fix!

PR #923, opened 2025-11-18
ci: drop vunel-specific qg runner

willmurphyscode:
There is a central definition of this runner over in the workflows repo. Drop the vunnel-specific runner of this name in favor of the central config.

The definition in vunnel is shadowing this one:

https://github.com/anchore/workflows/blob/1c6568eb1e45456271570411baedc438930b6d83/.github/runs-on.yml#L31-L45

Which had more space added. Use the one with more space.

Note that this change will have no effect until it's on `main` since that's how runs-on works.

### DIFF
--- a/.github/runs-on.yml
+++ b/.github/runs-on.yml
@@ -30,22 +30,6 @@ runners:
     spot: false
     extras: s3-cache
 
-  # quality-gate: beefy runners for vulnerability data processing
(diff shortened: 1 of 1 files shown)


willmurphyscode:
This seems not to have helped - https://github.com/anchore/vunnel/actions/runs/19473223739 eventually failed in the same place as the schedule runs.

PR #939, opened 2025-12-01
ci: run highly concurrent providers (ubuntu) on big runner

willmurphyscode:
The ubuntu provider does a ton of local file I/O and can be meaningfully sped up by using the a large multi core runner and a higher number of workers. However many providers are essentially single threaded and do not benefit from the multicore runner.

Therfore, rearrange some yaml to put highly concurrent providers (today just Ubuntu) on a bit multi core runner without paying to put the rest of them on the big multi core runner.

Requires: https://github.com/anchore/vunnel/pull/940

### DIFF
--- a/.github/workflows/nightly-quality-gate.yaml
+++ b/.github/workflows/nightly-quality-gate.yaml
@@ -5,13 +5,12 @@ on:
 
   # run 5 AM (UTC) daily
   schedule:
-    - cron:  '0 5 * * *'
+    - cron: "0 5 * * *"
 
 permissions:
   contents: read
 
 jobs:
-
   select-providers:
     runs-on: runs-on=${{ github.run_id }}/runner=small
     outputs:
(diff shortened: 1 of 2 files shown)


willmurphyscode:
Ubuntu run without this change: https://github.com/anchore/vunnel/actions/runs/19969982591/job/57272068623?pr=940

Ubuntu run with this change: https://github.com/anchore/vunnel/actions/runs/19970868274/job/57275079198?pr=939

Note that both runs have auto concurrency enabled.

src/vunnel/providers/rhel/csaf_client.py
```diff
@@ -54,15 +54,17 @@ def __init__(
         self,
         workspace: Workspace,
         logger: logging.Logger,
+        skip_download: bool,
+        max_workers: int,
         latest_url: str = ADVISORIES_LATEST_URL,
-        skip_download: bool = False,
```
spiffcs: Does this make an impact where this no longer has the default of `False` and we rely on the config being required to be passed?

src/vunnel/providers/rhel/rhsa_provider.py
```diff
@@ -196,17 +196,26 @@ class CSAFRHSAProvider(RHSAProvider):
     to use CSAF data as its source of RHSA data.
     """
 
-    def __init__(self, workspace: Workspace, download_timeout_seconds: int, logger: logging.Logger, skip_download: bool = False):
```
spiffcs: This one also removed the default for `skip_download`

PR #945, opened 2025-12-08
Use zstd-compressed fix date DBs

wagoodman:
This swaps which fix date archive that is being used to one that is using zstd compression. This fixes behavior where larger images are not being pulled and timing out.

We will attempt to use the `latest-zst` tag, unless it's unavailable, in which case we'll use the `latest` tag. In the future we can change the image used to be just `latest` explicitly.

### DIFF
--- a/src/vunnel/tool/fixdate/grype_db_first_observed.py
+++ b/src/vunnel/tool/fixdate/grype_db_first_observed.py
@@ -180,6 +180,7 @@ class Store(Strategy):
         if download_zst_path.exists():
             self.logger.debug(f"decompressing {download_zst_path} to {download_db_path}")
             dctx = zstandard.ZstdDecompressor()
+            # this will truncate or create any existing DBs and archives
             with download_zst_path.open("rb") as ifh, download_db_path.open("wb") as ofh:
                 dctx.copy_stream(ifh, ofh)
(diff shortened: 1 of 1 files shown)


src/vunnel/tool/fixdate/grype_db_first_observed.py
```diff
@@ -145,21 +146,63 @@ def _get_remote_digest(self, image_ref: str) -> str | None:
...
+        self.logger.info(f"pulling fix date database from {image_ref}")
+        client.pull(target=image_ref, outdir=str(download_dir))
+        self.logger.info(f"successfully fetched fix date database for {self.provider}")
+
+    def _process_downloaded_file(self, download_zst_path: Path, download_db_path: Path) -> None:
+        """decompress zstd file if present, otherwise verify db file exists."""
+        if download_zst_path.exists():
+            self.logger.debug(f"decompressing {download_zst_path} to {download_db_path}")
+            dctx = zstandard.ZstdDecompressor()
+            with download_zst_path.open("rb") as ifh, download_db_path.open("wb") as ofh:
```
willmurphyscode: Is there a path where the non-compressed `download_db_path` still exists from workspace cache of a previous run? If so, would this result in appending?

PR #949, opened 2025-12-11
Adding 0-deb Provider in Vunnel

raianand:
Hi, we at [KoalaLab](https://www.koalalab.com), have created a container-first distro 0-deb. I am raising this PR to add a Vunnel provider for 0-deb.

0-deb uses deb packaging and uses rolling release. Vulnerable range is used in security feed to indicate fixes for a particular package, CVE combination across different branches.


raianand:
Hi @willmurphyscode @wagoodman — thank you for the thoughtful review and for sharing the contributing guide and calendar link.
To answer your questions: yes, the registry and public test artifacts are still a work in progress on our end. Due to some priority shifts within our organization, we need to pause this effort for the next few weeks.
I'm closing this PR for now, but we remain committed to this integration. Once we have the registry live and can provide proper test cases (images demonstrating both false positives and negatives as you suggested), we'll either reopen this or submit a fresh PR.
Will definitely take you up on that calendar invite when we're ready to pick this back up. Thanks again for the guidance — it's genuinely helpful.

PR #959, opened 2025-12-19
fix: handle empty matrix jobs in quality gate evaluation

willmurphyscode:
Update nightly-quality-gate to use tag-based provider splitting (matching PR workflow). Fix evaluation logic in both workflows to accept "skipped" status, which GitHub Actions returns for empty matrices. Previously, the gate failed incorrectly when only multicore or only non-multicore providers needed testing.

https://github.com/anchore/vunnel/pull/958 has CI examples where it is failing even though both required providers succeeded.


wagoodman (approved):

wagoodman (approved):
posix compliant is better

PR #963, opened 2025-12-30
Add Root IO vulnerability data provider

chait-slim:
This adds a new provider for Root IO vulnerability data, which provides
  security information for Root IO patched packages across multiple
  ecosystems (Debian, Ubuntu, Alpine, NPM, PyPI).

  Implementation details:
  - Fetches OSV 1.6.1 format data from Root IO API (api.root.io/external/osv)
  - Implements NAK pattern: rootio- prefixed packages only match Root IO vulnerabilities
  - Supports ecosystem-specific version suffixes (.root.io.N for Debian/Ubuntu, -root.io.N for NPM, +root.io.N for PyPI)
  - Provider class in src/vunnel/providers/rootio/__init__.py
  - OSV record parser in src/vunnel/providers/rootio/parser.py
  - Registered in src/vunnel/providers/__init__.py and src/vunnel/cli/config.py


src/vunnel/providers/rootio/parser.py
```diff
@@ -0,0 +1,144 @@
...
+        """
+        # Fetch the list of OSV IDs
+        osv_ids = self._fetch_osv_ids()
+
+        # Download fixdate information if needed
+        # TEMPORARILY DISABLED: self.fixdater.download()
+        # Fix date patching is optional and requires authentication
+
+        # Fetch and process each OSV record
+        for osv_id in osv_ids:
```
willmurphyscode: Please make this concurrent in some way. Right now this provider does ~9K sequential, blocking http gets, which makes it very slow for a relatively small amount of data. Many of the other providers have some `concurrent.futures.ThreadPoolExecutor` use and a config that controls the concurrency (and sets a default higher than 1). Please imitate that pattern here.

willmurphyscode: It's probably fine to enter a concurrent section that pulls down all the osv docs and then process them sequentially, which is probably easier than trying to get the entire record normalized and processed concurrently.

chait-slim: Added concurrency

tests/quality/config.yaml
```diff
@@ -374,3 +374,35 @@ tests:
...
+      - name: alpine
+        use_cache: true
+      - name: debian
+        use_cache: true
+      - name: ubuntu
+        use_cache: true
+      - name: github
+        use_cache: true
+    images:
+      - cr.root.io/cassandra@sha256:b3cc918a6a364af0a6b0a45becef0d0979db7e604751fad627ec2a94945b4e03
```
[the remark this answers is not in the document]
chait-slim: Fixed

chait-slim:
@willmurphyscode Following our talk today:
1. grype repo - updated code now passes static analysis and unit tests locally
2. vunnel repo - rebased
3. labels repo - updated image and validated quality gate locally

src/vunnel/providers/rootio/parser.py
```diff
@@ -0,0 +1,193 @@
...
+        """Fetch the list of OSV record IDs from the Root IO API."""
+        self.logger.info("fetching list of OSV IDs from Root IO")
+        url = f"{self.api_base_url}/all.json"
+        response = http.get(url, self.logger, timeout=self.download_timeout)
+
+        # Parse the response - it's an array of objects with "id" and "modified" fields
+        id_objects = response.json()
+
+        # Extract and validate ID strings from each object
+        all_ids = [obj["id"] for obj in id_objects]
```
willmurphyscode: In local testing, I had to add `.strip()` here because some ids in all.json have a space on the end. Can you do that here (and maybe fix all.json as well)?

tests/quality/config.yaml
```diff
@@ -25,7 +25,7 @@ yardstick:
       # Note:
       #  - ALWAYS leave the "import-db" annotation as-is
       #  - this version should ALWAYS match that of the other "grype" tool below
-      version: main+import-db=build/vulnerability.db
+      version: github.com/chait-slim/grype@feat/rootio-support+import-db=build/vulnerability.db
```
willmurphyscode: I was trying to suggest that this be changed for local testing, not necessarily that the changes needed to be pushed. For example when you run the quality gate test you can do `file:///path/to/grype+...` here and on line 39 below, and to a local checkout of `grype-db` that has a `go work use . ../grype` in the grype db version.

Running the tests like this doesn't include database build time changes (because `grype-db@main` doesn't have your change from the grype PR yet), so when you run the quality gate with the config file as is, you only see differences in canonical's upstream data between when the reference db was published and when the cache was built, and don't see the effect of an actually changed rootio matching. (We're working to improve this part; I know it's confusing.)

chait-slim: fixed

PR #968, opened 2026-01-08
feat: arch linux provider

willmurphyscode:


tests/unit/providers/arch/test_arch.py
wagoodman: I think this is missing schema conformance tests -- see the other provider unit tests (they tend to be the last ones in the test file)

PR #1136, opened 2026-04-16
feat(ubuntu): emit fixed ins for Not Affected records

willmurphyscode:
Previously, the ubuntu provider would drop records where upstream data indicated that a package was not affected by a CVE; however, that leaves downstream consumers unable to distinguish between a search miss (no record of that CVE affecting that package) and a true "not affected" (the vendor examined their code and decided the vulnerability did not apply).

Instead, emit a FixedIn record that essentially says, "this was fixed at version 0" in situations where there upstream data indicates something is not affected.

### DIFF
--- a/src/vunnel/providers/ubuntu/parser.py
+++ b/src/vunnel/providers/ubuntu/parser.py
@@ -556,18 +556,37 @@ def map_parsed(parsed_cve: CVEFile, fixdater: fixdate.Finder, logger: logging.Lo
             r.NamespaceName = namespace_name
             vulns[namespace_name] = r
 
+        # Emit an explicit "not affected" FixedIn (version "0") so downstream consumers
+        # can distinguish "confirmed not affected" from "no data available".
+        if p.status == "not-affected":
+            pkg = FixedIn()
+            pkg.Name = p.package
+            pkg.Version = "0"
+            pkg.VendorAdvisory = {"NoAdvisory": False}
+            pkg.VersionFormat = "dpkg"
+            pkg.NamespaceName = namespace_name
+            r.FixedIn.append(pkg)
+            continue
+
         # If the patch status is one we care about, make the FixedIn record, else skip it but create CVE records
         # We currently want to mark end-of-support records with no previously known fix as vulnerable, hence the
         # or check_merge step here.
         if check_state(p.status) or check_merge(p):
             # If the patch is needs-triage but a corresponding ESM entry confirms
-            # the package is not affected, skip emitting a FixedIn record. This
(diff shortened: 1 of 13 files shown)


src/vunnel/providers/ubuntu/parser.py
```diff
@@ -556,18 +556,37 @@ def map_parsed(parsed_cve: CVEFile, fixdater: fixdate.Finder, logger: logging.Lo
...
         # or check_merge step here.
         if check_state(p.status) or check_merge(p):
             # If the patch is needs-triage but a corresponding ESM entry confirms
-            # the package is not affected, skip emitting a FixedIn record. This
-            # prevents false-positive match-all constraints for packages where
-            # Ubuntu's ESM review determined the vulnerable code is not present.
+            # the package is not affected, emit an explicit "not affected" FixedIn
+            # (version "0") instead of the match-all constraint that needs-triage
+            # would normally produce.
             if p.status == "needs-triage" and (p.distro, p.package) in esm_not_affected:
```
kzantow, line 579: I this means: ubuntu has a needs-triage and esm indicates not affected, we are assuming that the main record is not affected. 👍 

But are there more not affected records in `ignored_patches` we should be including?

willmurphyscode: IME when canonical says, "ignored" they mean, "we aren't going to fix this" not "we aren't affected," but I'm not sure what you mean by `ignored_matches` here.

kzantow: The esm_not_affected comes from `for ip in parsed_cve.ignored_patches:` it seems like there are records in the `.ignore_patches` that are not in `.patches` that maybe should be included

willmurphyscode: I think I understand your question now. `ignore_patches` are patches that we can't find a namespace for: https://github.com/anchore/vunnel/blob/80bf474d193fbae78823746f3a540a7f7947f991/src/vunnel/providers/ubuntu/parser.py#L989 at https://github.com/anchore/vunnel/blob/80bf474d193fbae78823746f3a540a7f7947f991/src/vunnel/providers/ubuntu/parser.py#L965

So generally these are records we don't care about not because they are an unaffected package we're skipping, but because they apply to a _distro version_ we're skipping. The ESM ones are a special case where, if Ubuntu said something was unaffected during the ESM period, we're comfortable assuming it was unaffected _before_ too (because there's no ESM specific patch so it's the same code.)

I think that means there aren't any other ones we need to look for as part of this PR.

kzantow (approved):

PR #1137, opened 2026-04-16
feat(debian): emit fixed ins for version 0 records

willmurphyscode:
Previously, the Debian provider would drop records where upstream data indicated a package was fixed at "version 0", since these records could never result in a vulnerability match in grype. However, this left downstream providers unable to distinguish between a "search miss" (upstream data has no mention of this CVE/package/distro version tuple) and an explicit indication of "not affected" (Debian maintainers believe this CVE/package/distro version tuple is not affected).

Instead, forward the idea of "fixed at version 0" into the results, so that downstream consumers can distinguish between a search miss and a package that maintainers believe is not affected.

### DIFF
--- a/src/vunnel/providers/debian/parser.py
+++ b/src/vunnel/providers/debian/parser.py
@@ -387,11 +387,16 @@ class Parser:
                             if "fixed_version" in distro_record:
                                 fixed_el["Version"] = distro_record["fixed_version"]
                                 if distro_record["fixed_version"] == "0":
-                                    # version == 0 should mean that the
-                                    # package was determined to not be
-                                    # vulnerable in the distro namespace
-                                    # (from reviewing
-                                    # https://security-tracker.debian.org/tracker/)
+                                    # version == 0 means the package was determined
+                                    # to not be vulnerable in this distro namespace.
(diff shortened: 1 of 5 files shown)


src/vunnel/providers/debian/parser.py
```diff
@@ -387,11 +387,16 @@ def _normalize_json(self, ns_cve_dsalist=None):  # noqa: PLR0912,PLR0915,C901
                             if "fixed_version" in distro_record:
                                 fixed_el["Version"] = distro_record["fixed_version"]
                                 if distro_record["fixed_version"] == "0":
-                                    # version == 0 should mean that the
-                                    # package was determined to not be
-                                    # vulnerable in the distro namespace
-                                    # (from reviewing
-                                    # https://security-tracker.debian.org/tracker/)
+                                    # version == 0 means the package was determined
```
kzantow, line 390: Am I reading this correctly that we were including these records already, just with no fixed-in data, so it was not applying to anything? That's encouraging about db size! :)

kzantow (approved):
LGTM but I assume this needs to be tested against current grype main for quality gate passing?

PR #1205, opened 2026-05-27
add chainguard osv feed

crosleyzack:
### What

Adds the ability for the chainguard provider to consume the Chainguard OSV feed in addition to the chainguard secdb feed

### Why

We are trying to migrate to OSV due to the additional context and fidelity it provides. The SecDB feed is slated to be deprecated end of this year (2026)

### Notes

The quality tests passed, which is to be expected as no behavior should have changed.

```
✦2 ❯ make validate provider=chainguard
uv run yardstick validate --result-set pr_vs_latest_via_sbom_chainguard
Loading label entries...done! 23 entries loaded
Validating with 'pr_vs_latest_via_sbom_chainguard'
2026-05-27 16:47:07,206 [INFO] only considering matches from allowed namespaces: chainguard:distro:chainguard:rolling
2026-05-27 16:47:07,206 [INFO] Testing image: 'ghcr.io/chainguard-images/scanner-test@sha256:59bddc101fba0c45d5c093575c6bc5bfee7f0e46ff127e6bb4e5acaaafb525f9' with 'syft@latest, grype@main+import-db=build/vulnerability.db, grype@main+import-db=https://grype.anchore.io/databases/v6/vulnerability-db_v6.1.4_2026-05-13T00:47:21Z_1778657102.tar.zst'
   Results used for image ghcr.io/chainguard-images/scanner-test@sha256:59bddc101fba0c45d5c093575c6bc5bfee7f0e46ff127e6bb4e5acaaafb525f9:
    ├── c9988b12-2a36-4fb4-9490-db690002220d : grype[custom-db]@v0.99.1-14-g8a04a093 (custom-db)  against ghcr.io/chainguard-images/scanner-test@sha256:59bddc101fba0c45d5c093575c6bc5bfee7f0e46ff127e6bb4e5acaaafb525f9
    └── ddec88a3-940d-47ec-b954-8541a59680e3 : grype[reference]@v0.99.1-14-g8a04a093 (reference)  against ghcr.io/chainguard-images/scanner-test@sha256:59bddc101fba0c45d5c093575c6bc5bfee7f0e46ff127e6bb4e5acaaafb525f9
--------------------------------------------------------------------------------

Quality gate passed!
```

Succesfully ran `VUNNEL_PROVIDERS_CHAINGUARD_USE_OSV=true VUNNEL_ROOT=data_osv uv run vunnel run chainguard` and `GRYPE_DB_PROVIDER_ROOT=../vunnel/data_osv GRYPE_DB_BUILD_DIR=./db_osv/6 go run ./cmd/grype-db build --schema 6` to create grype-db

Part of Chainguard issue INT-520


src/vunnel/providers/wolfi/parser.py
```diff
@@ -171,19 +195,190 @@ def _normalize(self, release, data):  # noqa: C901
...
+                                    "ComponentName": component["component_name"],
+                                    "ComponentVersion": component["component_version"],
+                                    "ComponentType": component["component_type"],
+                                    "ComponentLocation": component["component_location"],
+                                    "ComponentPurl": component["component_purl"],
+                                    "LatestEventStatus": component["latest_event_status"],
+                                    "LatestEventTimestamp": component["latest_event_timestamp"],
+                                }
+                                for component in affected["ecosystem_specific"]["components"]
+                            ]
```
crosleyzack: This information is pulled from our environment specific data for our [osv v2 feed](https://packages.cgr.dev/chainguard/v2/osv/CGA-2255-2h2p-73q2.json). We have an open question of how we should format this information for grype, namely:
1. Should the vulnerability entries be generated at this lower component level
2. Where can this additional information be added
3. How do we store and use this information in grypedb and grype respectively

src/vunnel/providers/wolfi/parser.py
```diff
@@ -171,19 +195,190 @@ def _normalize(self, release, data):  # noqa: C901
...
+        ecosystem = f"{self.namespace}:{release}"
+        cvss_scores = [
+            _parse_cvss(entry["score"])
+            for entry in data["severity"]
+        ]
+
+        return {
+            "Vulnerability": {
+                "CVSS": cvss_scores,
+                "Description": "",
```
crosleyzack: Should we autogenerate this field or drop it

src/vunnel/providers/wolfi/parser.py
```diff
@@ -171,19 +195,190 @@ def _normalize(self, release, data):  # noqa: C901
...
+                                for component in affected["ecosystem_specific"]["components"]
+                            ]
+                        },
+                        "Version": _version_from_ranges(affected["ranges"]),
+                        "VersionFormat": "apk",
+                        # TODO should we try to pull timestamp from components
+                        # "Availability": {
+                        #     "Date": ,
+                        #     "Kind": "advisory",
+                        # },
```
crosleyzack: This doesn't appear to be a required field, can we omit this?

src/vunnel/providers/chainguard/__init__.py
```diff
@@ -23,9 +23,15 @@ class Config:
...
+    osv_url: str = "https://packages.cgr.dev/chainguard/v2/osv/all.json"
+    # Override with VUNNEL_PROVIDERS_CHAINGUARD_USE_OSV
+    use_osv: bool = False
 
 
 class Provider(provider.Provider):
+    # NOTE: schema and distribution version are actually set on init depending
+    # on which feed we configure the provider to use.
     __schema__ = schema.OSSchema()
     __distribution_version__ = int(__schema__.major_version)
```
crosleyzack, line 37: Is it a problem if this always reports OSSchema? I ensured the argument to `writer.write` below is the actual, correct schema

crosleyzack:
This appears to be a dupe of [a PR by Ville](https://github.com/anchore/vunnel/pull/1082/) updating the vunnel provider. Humbly, I think creating a new parser is much cleaner as an approach though

crosleyzack:
`VUNNEL_PROVIDERS_CHAINGUARD_USE_OSV=true VUNNEL_ROOT=data_osv uv run vunnel run chainguard` successfully creates vunnel output
`GRYPE_DB_PROVIDER_ROOT=../vunnel/data_osv GRYPE_DB_BUILD_DIR=./db_osv/6 go run ./cmd/grype-db build --schema 6` successfully creates grype-db

src/vunnel/providers/wolfi/parser.py
```diff
@@ -171,19 +253,90 @@ def _normalize(self, release, data):  # noqa: C901
...
+            index = orjson.loads(r.content)
+
+            base_url = self.url.rsplit("/", 1)[0]
+            # Download all entries in the index concurrently using a thread pool,
+            # which should speed up the download process significantly since there are thousands of entries.
+            # We construct the URL for each entry by appending the entry ID and .json to the base URL
+            # e.g. https://packages.cgr.dev/chainguard/v2/osv/CGA-2255-2h2p-73q2.json
+            with concurrent.futures.ThreadPoolExecutor(max_workers=self.max_workers) as executor:
+                futures = [
+                    executor.submit(self._download_single_file, f"{base_url}/{entry['id']}.json", f"{entry['id']}.json")
```
willmurphyscode: `entry['id']` could use some validation, since right now it will be `path.join`ed and came from the network. Simple regex is probably fine.

src/vunnel/providers/wolfi/parser.py
```diff
@@ -171,19 +253,90 @@ def _normalize(self, release, data):  # noqa: C901
...
+
+    def _download_single_file(self, url: str, filename: str) -> None:
+        '''
+        Download a single OSV entry file given its URL and the desired filename.
+        '''
+        file_path = os.path.join(self.input_dir_path, filename)
+        # if the file already exists and skip_redownload is True, skip writing the file again. This is to avoid
+        # unnecessary redownloading and rewriting of the same file, which can save time on subsequent
+        # runs.
+        if self.skip_redownload and os.path.exists(file_path):
```
willmurphyscode: What's the point of `skip_redownload`? Is there a reason we can't get to "download if changed but always"? I don't like giving the user switches where one of the switches does the wrong thing.

Providers _generally_ have a `--skip-download` which is useful in testing / re-runnign on existing workspace state. Is there a reason we need a separate `skip-redownload`?

src/vunnel/providers/wolfi/parser.py
```diff
@@ -171,19 +253,90 @@ def _normalize(self, release, data):  # noqa: C901
...
+            # https://packages.cgr.dev/chainguard/v2/osv/all.json
+            r = http.get(self.url, self.logger, timeout=self.download_timeout)
+            index = orjson.loads(r.content)
+
+            base_url = self.url.rsplit("/", 1)[0]
+            # Download all entries in the index concurrently using a thread pool,
+            # which should speed up the download process significantly since there are thousands of entries.
+            # We construct the URL for each entry by appending the entry ID and .json to the base URL
+            # e.g. https://packages.cgr.dev/chainguard/v2/osv/CGA-2255-2h2p-73q2.json
+            with concurrent.futures.ThreadPoolExecutor(max_workers=self.max_workers) as executor:
```
willmurphyscode: Are you all opening to publishing a zip / targz of the whole set? Some providers do that and it saves a _ton_ of network connections.

willmurphyscode: A common pattern is basically like `snapshot.tar.gz` and `changes.csv` which is a list of things that changes since the snapshot, if it can't be regenerated.

src/vunnel/providers/wolfi/parser.py
```diff
@@ -171,19 +253,90 @@ def _normalize(self, release, data):  # noqa: C901
...
+                    continue
+                self.logger.info(f"loading {self.namespace} osv data from {filename}")
+                with open(os.path.join(self.input_dir_path, filename)) as fh:
+                    data = orjson.loads(fh.read())
+                    yield self._release_, data
+        except Exception:
+            self.logger.exception(f"failed to load {self.namespace} osv data")
+            raise
+
+    def _normalize(self, release: str, data: dict[str, Any]) -> dict[str, Any]:  # noqa: C901
```
willmurphyscode: Does your feed include fixed dates? (Dates when fixed versions were published?) If not, this is a good place to call https://github.com/anchore/vunnel/blob/a87bb48403dda4f7447ecbecc59940d1dd2da04d/src/vunnel/utils/osv.py#L7 I think (unless you're calling it somewhere else and I missed it.)

src/vunnel/providers/wolfi/parser.py
crosleyzack: I didn't know that `skip-download` existed and added this for testing -- I can remove it

src/vunnel/providers/wolfi/parser.py
crosleyzack: We do not currently, but can look into that. Another coworker has messed with the multithreaded download and improved its performance substantially

src/vunnel/providers/wolfi/parser.py
willmurphyscode: Yep providers need to opt in:

https://github.com/anchore/vunnel/blob/a87bb48403dda4f7447ecbecc59940d1dd2da04d/src/vunnel/providers/rhel/__init__.py#L65-L83 for example

src/vunnel/providers/wolfi/parser.py
crosleyzack: added!

src/vunnel/providers/wolfi/parser.py
crosleyzack: set it up to use the built in `skip-download`

src/vunnel/providers/wolfi/parser.py
crosleyzack: validated against CGA regex

src/vunnel/providers/wolfi/parser.py
```diff
@@ -171,19 +252,72 @@ def _normalize(self, release, data):  # noqa: C901
...
+            self.logger.info(f"skip_download is enabled for {self.namespace} osv feed")
+            return
+
+        if not os.path.exists(self.input_dir_path):
+            os.makedirs(self.input_dir_path, exist_ok=True)
+
+        try:
+            self.logger.info(f"downloading {self.namespace} osv {self.url}")
+            r = http.get(self.url, self.logger, stream=True, timeout=self.download_timeout)
+            buf = io.BytesIO(r.content)
```
willmurphyscode: This downloads via stream and then immediately reads the whole stream. I think you probably want to stream to the file like https://github.com/anchore/vunnel/blob/9256447f49270ea3c7ded5ff1c288987c49ced60/src/vunnel/providers/rhel/csaf_client.py#L100

src/vunnel/providers/wolfi/parser.py
```diff
@@ -171,19 +252,72 @@ def _normalize(self, release, data):  # noqa: C901
...
+                filenames.append(filename)
+
+            # given the large volume of osv files, process in parallel
+            def _load_file(filename: str) -> dict[str, Any]:
+                self.logger.debug(f"loading {self.namespace} osv data from {filename}")
+                with open(os.path.join(self.input_dir_path, filename)) as fh:
+                    return orjson.loads(fh.read())
+
+            with concurrent.futures.ThreadPoolExecutor(max_workers=self.max_workers) as executor:
+                for data in executor.map(_load_file, filenames):
```
willmurphyscode: I think map is eager, see https://docs.python.org/3.13/library/concurrent.futures.html#concurrent.futures.Executor.map , and this results in a slot in the map per file available being eagerly loaded. Maybe we don't need concurrency here any more since this is local I/O? Or maybe a pattern can limit the concurrency?

Honestly I think we can just skip extracting and yield from a decompression string like `with tarfile.open(self._archive_path, mode="r|gz") as tf:`.

I think the eager map is making this really slow - local testing shows a naive stream from tar.gz on disk is much faster.

src/vunnel/providers/wolfi/__init__.py
```diff
@@ -24,7 +24,9 @@ class Config:
     request_timeout: int = 125
     # Override with VUNNEL_PROVIDERS_WOLFI_SECDB_URL
     secdb_url: str = "https://packages.wolfi.dev/os/security.json"
-
+    # Override with VUNNEL_PROVIDERS_WOLFI_ENABLE
+    # Enable allows us to switch to an OSV feed in the future if/when one becomes available
+    enable: bool = True
```
willmurphyscode, line 29: Why does this have a different name for Wolfi than for Chainguard?

src/vunnel/providers/wolfi/__init__.py
```diff
@@ -60,6 +62,10 @@ def tags(cls) -> list[str]:
         return ["vulnerability", "os"]
 
     def update(self, last_updated: datetime.datetime | None) -> tuple[list[str], int]:
+        if not self.config.enable:
```
willmurphyscode, line 71: I don't understand this. Why do we have a control flow where we make this no op but still call it?

src/vunnel/providers/wolfi/parser.py
```diff
@@ -171,19 +248,73 @@ def _normalize(self, release, data):  # noqa: C901
...
+        https://github.com/anchore/vunnel/blob/main/schema/vulnerability/osv/schema-1.7.0.json
+
+        Input shape reference (Chainguard OSV):
+        https://packages.cgr.dev/chainguard/v2/osv/CGA-2255-2h2p-73q2.json
+        """
+        # we map the osv id to the osv data to keep consistency in the secdb parser, which
+        # does this for ease of identifying the associated vulnerability when writing records.
+        # IE: {"CGA-1234-5678-9abc": {<full osv record>}}
+        osv.patch_fix_date(data, self.fixdater)
+        return {data["id"]: data}
```
willmurphyscode: If there's a single JSON in the tar with no `id` key, this will cause the whole parser to exit. Probably worth a `data.get("id")` and a `log.warn ; continue` or similar if there's nothing there?

src/vunnel/providers/wolfi/parser.py
```diff
@@ -171,19 +248,73 @@ def _normalize(self, release, data):  # noqa: C901
...
+                    if not n.endswith(".json"):
+                        continue
+                    if not self._cga_id_re.match(n.removesuffix(".json")):
+                        continue
+                    f = tf.extractfile(member)
+                    if not f:
+                        continue
+
+                    yield self._release_, orjson.loads(f.read())
+        except Exception:
```
willmurphyscode: Can we catch a narrower class of exception? What do we expect to go wrong here?

src/vunnel/providers/wolfi/parser.py
```diff
@@ -171,19 +248,73 @@ def _normalize(self, release, data):  # noqa: C901
...
-        # load the data
-        for release, dbtype_data_dict in self._load():
-            # normalize the loaded data
-            yield release, self._normalize(release, dbtype_data_dict)
+        try:
+            # download the tar file
+            self.logger.info(f"downloading {self.namespace} osv {self.url}")
+            self._download_stream(self.url, os.path.join(self.input_dir_path, self._tar_file))
+        except Exception:
+            self.logger.exception(f"ignoring error downloading osv feed for {self.url}")
```
willmurphyscode, line 279: Do we want to serve stale data if the stream fails to download?

src/vunnel/providers/wolfi/parser.py
```diff
@@ -171,19 +248,73 @@ def _normalize(self, release, data):  # noqa: C901
...
+    _input_dir_ = "osv"
+    _cga_id_re  = re.compile(r"^CGA(-[23456789cfghjmpqrvwx]{4}){3}$")
+    _tar_file   = "chainguard.tar.gz"
+
+    def _download(self) -> None:
         """
-        Download, load and normalize wolfi sec db and return a dict of release - list of vulnerability records
-        :return:
+        Download all OSV entry files based on the index file at self.url, which should point to the
+        top level all.json file. For each entry in the index, we construct the URL for the individual
```
willmurphyscode: Needs to be updated now that we're using a `.tar.gz`.

src/vunnel/providers/wolfi/parser.py
```diff
@@ -171,19 +248,73 @@ def _normalize(self, release, data):  # noqa: C901
...
+            os.makedirs(self.input_dir_path, exist_ok=True)
 
-        # load the data
-        for release, dbtype_data_dict in self._load():
-            # normalize the loaded data
-            yield release, self._normalize(release, dbtype_data_dict)
+        try:
+            # download the tar file
+            self.logger.info(f"downloading {self.namespace} osv {self.url}")
+            self._download_stream(self.url, os.path.join(self.input_dir_path, self._tar_file))
```
willmurphyscode: Should this have a timeout?

src/vunnel/providers/wolfi/parser.py
```diff
@@ -171,19 +248,73 @@ def _normalize(self, release, data):  # noqa: C901
...
+    def _load(self) -> Generator[tuple[str, dict[str, Any]]]:
+        self.logger.info(f"load files for {self.namespace} osv feed")
+        try:
+            with tarfile.open(os.path.join(self.input_dir_path, self._tar_file), mode="r:gz") as tf:
+                # Extract a specific file as a file-like object
+                for member in tf:
+                    n = member.name
+                    if not n.endswith(".json"):
+                        continue
+                    if not self._cga_id_re.match(n.removesuffix(".json")):
```
willmurphyscode, line 296: Previous implementation logged a warning here. Should this still do so?

src/vunnel/providers/wolfi/__init__.py
crosleyzack: Because the Wolfi information is in the same OSV feed as the chainguard info, so to have wolfi use the chainguard information it would be `use_osv == true` in Chainguard and disabling wofli. This is part of our larger move to no longer have multiple namespaces for our OS

src/vunnel/providers/wolfi/__init__.py
crosleyzack: Because the wolfi information is now in the same OSV feed as the chainguard info. Once chainguard feed is using OSV, the wolfi provider is no longer necessary

src/vunnel/providers/wolfi/parser.py
crosleyzack: That is a good point, weirdly this is also what secdb seems to be doing which feels wrong

src/vunnel/providers/wolfi/parser.py
crosleyzack: fixed in 85d09248e8823e81db2680f91427242af4b31b84

src/vunnel/providers/wolfi/parser.py
crosleyzack: fixed in 85d09248e8823e81db2680f91427242af4b31b84

src/vunnel/providers/wolfi/parser.py
crosleyzack: fixed in 85d09248e8823e81db2680f91427242af4b31b84

src/vunnel/providers/wolfi/parser.py
crosleyzack: fixed in 85d09248e8823e81db2680f91427242af4b31b84

src/vunnel/providers/wolfi/parser.py
crosleyzack: fixed in 85d09248e8823e81db2680f91427242af4b31b84

src/vunnel/providers/wolfi/parser.py
crosleyzack: fixed in 85d09248e8823e81db2680f91427242af4b31b84

src/vunnel/providers/wolfi/__init__.py
willmurphyscode: What should the distro field be? I think some images have `wolfi` in /etc/os-release and some have `chainguard`. Is that going to stop happening?

tests/unit/providers/chainguard/test_chainguard.py
```diff
@@ -1,16 +1,73 @@
 from __future__ import annotations
 
+import io
```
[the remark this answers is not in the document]
willmurphyscode: import removed.

src/vunnel/providers/wolfi/__init__.py
willmurphyscode: My understanding is this: at some point, Wolfi will stop being it's own namespace (all the data will be in the chainguard feed), and this gives us a config-only switch to turn off the wolfi feed when that happens. makes sense and looks good.

willmurphyscode (approved):

src/vunnel/providers/wolfi/__init__.py
crosleyzack: We still have both. If you look at https://packages.cgr.dev/chainguard/v3/osv/CGA-27xx-xp52-rjpm.json as an example, we have two affected entries: one for Chainguard and one for Wolfi

PR #1206, opened 2026-05-28
fix: clean destination prior to extracting archive

westonsteimel:
Otherwise artifacts from previous caches that have been removed in upstream data will continue to be considered in the output


westonsteimel:
Once released the next published grype db should clear up https://github.com/anchore/grype/issues/3456

wagoodman (approved):

PR #1212, opened 2026-05-29
Switch Ubuntu provider to use OSV feed

willmurphyscode:
The Ubuntu provider previously worked by cloning down (or doing a `git pull` in a cached dir) the ubuntu-cve-tracker git repo. However, this repo is prohibitively large and not really meant to be user facing. We've also had operational difficulties where even on very large VMs, processing major updates could take 7+ hours.

Therefore, switch to the OSV feed, which has other limitations but is at least available as a single, current archive that is operationally simpler to work with. This PR adds a README.md file for the Ubuntu provider, which shares details about the implementation and decisions in it.


wagoodman (approved):

src/vunnel/providers/ubuntu/parser.py
```diff
@@ -1,1203 +1,689 @@
...
+    inference provenance.
+    """
+    new_affs: list[dict[str, Any]] = []
+    for aff in pro_affs:
+        pkg = aff.get("package", {}).get("name")
+        if not pkg or pkg in existing_pkgs:
+            continue
+        existing_pkgs.add(pkg)
+        synth = _build_synthetic_base_affected(aff, base_eco)
+        synth["database_specific"]["anchore"]["inference"] = {
```
willmurphyscode, line 122: This is trying to say, "this record is not in the OSV data as is, but we infer its existence from the ESM data." It's very similar to what we're doing here: https://github.com/anchore/vunnel/pull/1077

The idea is: this wasn't triaged before Ubuntu 14.04 went EOL, but when the triaged it during Ubuntu 14.04 Pro, it was found to be vulnerable, so we infer that it was always vulnerable.

PR #1214, opened 2026-06-01
feat(rhel): use version ranges for multi upstream RHSAs

### LINKED ISSUE #1195: Red Hat RHSA missing overlap
There is a bizarre case with RHSA data we are incorrectly handling

The example advisory we will use is
https://access.redhat.com/security/cve/cve-2024-8088
specifically python 3.9 on RHEL9. There are two python 3.9 advisories for RHEL9

Today we use the greatest version and report anything less as vulnerable. If we look at what Grype has to say about this CVE we see

```
CVE-2024-8088  python3.9    rpm     redhat:distro:redhat:9     < 0:3.9.19-8.el9
```

But if we look at the RHSAs, we see this CVE is also fixed in version `3.9.18-3.el9_4.5` which is less than `3.9.19-8.el9`

So this fixed version will be marked as vulnerable even when it is not vulnerable

We don't want to set the affected versions to `< 3.9.18-3.el9_4.5` because it is possible there could be a vulnerable version of a package released between the two versions of this sort of advisory (I didn't go looking for an example but after a conversation with Red Hat they confirmed it is theoretically possible)

# Possible solutions

The easiest solution to this would be to mark the singular version `3.9.18-3.el9_4.5` as not affected and keep the current range in place

There is probably a more complex and complete solution to this, but it's not clear to me how that would work.

willmurphyscode:
When Red Hat ships more than one fix for the same package in a single RHEL major version at distinct upstream version bases — a minor-stream backport that keeps the older base (release tag .elN_M) plus a later rebase to a newer base (release tag .elN) — Hydra reports both under the same "Red Hat Enterprise Linux N" product. Previously the parser collapsed them to the single highest fix, falsely flagging hosts that already carry the older stream's backport (e.g. python3.9-3.9.18-3.el9_4.5 flagged against the 3.9.19-8.el9 fix for CVE-2024-8088, because 3.9.18 < 3.9.19 in RPM ordering).

_parse_affected_release now collects all fixes per (package, platform, module), reduces them to distinct upstream bases, and when more than one base exists emits a VulnerableRange that partitions each stream below its own fix:

    < 0:3.9.18-3.el9_4.5 || >= 0:3.9.19, < 0:3.9.19-8.el9

grype-db uses VulnerableRange verbatim and grype already evaluates the compound constraint, so existing clients match correctly with no upgrade. The reported Version stays the newest stream's fix (single-constraint fallback) and all contributing RHSAs are folded into the advisory summary.

The change is additive: single-stream and same-base groups behave exactly as before. Same-base streams (only the .elN_M dist tag differs, e.g. .el9_2 vs .el9_4 of glibc 2.34) can't be separated by RPM version comparison and remain for a later change.

See https://access.redhat.com/security/cve/cve-2024-8088#cve-affected-packages `python3.9` for RHEL 9 for an example.

```
$ grype db search --vuln  CVE-2024-8088 --distro rhel:9
VULNERABILITY  PACKAGE     ECOSYSTEM  NAMESPACE               VERSION CONSTRAINT
CVE-2024-8088  python3.11  rpm        redhat:distro:redhat:9  < 0:3.11.9-7.el9
CVE-2024-8088  python3.12  rpm        redhat:distro:redhat:9  < 0:3.12.5-2.el9
CVE-2024-8088  python3.9   rpm        redhat:distro:redhat:9  < 0:3.9.18-3.el9_4.5 || >= 0:3.9.19, < 0:3.9.19-8.el9
```

### DIFF
--- a/src/vunnel/providers/rhel/parser.py
+++ b/src/vunnel/providers/rhel/parser.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import concurrent.futures
 import contextlib
 import copy
+import functools
 import logging
 import os
 import queue
@@ -35,7 +36,11 @@ if TYPE_CHECKING:
 namespace = "rhel"
 
 
-FixedIn = namedtuple("FixedIn", ["package", "platform", "version", "module", "advisory"])
+FixedIn = namedtuple(
+    "FixedIn",
+    ["package", "platform", "version", "module", "advisory", "vulnerable_range", "additional_advisories"],
+    defaults=[None, ()],
+)
 RHSAFixedIn = namedtuple("RHSAFixedIn", ["package", "version"])
 Advisory = namedtuple("Advisory", ["wont_fix", "rhsa_id", "link", "severity"])
 NamespacePayload = namedtuple("NamespacePayload", ["namespace", "payload"])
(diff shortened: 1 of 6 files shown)


spiffcs:
> _parse_affected_release now collects all fixes per (package, platform, module), reduces them to distinct upstream bases, and when more than one base exists emits a VulnerableRange that partitions each stream below its own fix:
😮 

Very cool! Taking a look at the larger review now and will 🟢 if no questions 😄

spiffcs (approved):

PR #1222, opened 2026-06-16
chore: turn down rhel OVAL mode

willmurphyscode:
The RHEL provider has been happily running in production in CSAF mode for some time. Remove the OVAL parser.

Then, since the utils.oval_parser had no other clients besides the oracle provider, move it into the oracle provider to keep all the oracle code together, and to discourage adding new providers that depend on the oval_parser.


westonsteimel (approved):

wagoodman (approved):
borderline cathartic

PR #1232, opened 2026-06-25
feat(RHEL): handle multiple fix streams

willmurphyscode:
Previously, when there were multiple RHSAs that fixed the same RPM for the same RHEL major version, Vunnel would either collapse them into a complex boolean expression if possible (for example if upstream moved and EVR was enough to distinguish the versions) or else would collapse to the higher version to prevent false negatives.

Instead, emit an additional item in the JSON for each additional relevant advisory. Crucially, the highest EVR is still in the "version" field so that legacy consumers of the data won't know the change, but these additional advisories will enable new consumers to decide which RHSA is most applicable to their stream.

Note that this makes the 'elN_M' suffixes on RPM versions somewhat load bearing, since that is how stream information is reported through, but that convention seems very strong and well-established.

### DIFF
--- a/tests/unit/providers/rhel/test-fixtures/snapshots/rhel:6/cve-2017-3511.json
+++ b/tests/unit/providers/rhel/test-fixtures/snapshots/rhel:6/cve-2017-3511.json
@@ -18,6 +18,15 @@
       "Description": "An untrusted library search path flaw was found in the JCE component of OpenJDK. A local attacker could possibly use this flaw to cause a Java application using JCE to load an attacker-controlled library and hence escalate their privileges.",
       "FixedIn": [
         {
+          "AdditionalAdvisoryFixes": [
+            {
+              "Advisory": {
+                "ID": "RHSA-2017:1109",
+                "Link": "https://access.redhat.com/errata/RHSA-2017:1109"
+              },
+              "Version": "1:1.8.0.131-0.b11.el6_9"
+            }
+          ],
           "Available": {
             "date": "2024-01-01",
             "kind": "first-observed"
@@ -38,6 +47,15 @@
           "VersionFormat": "rpm"
         },
         {
+          "AdditionalAdvisoryFixes": [
+            {
+              "Advisory": {
+                "ID": "RHSA-2017:1204",
+                "Link": "https://access.redhat.com/errata/RHSA-2017:1204"
+              },
+              "Version": "1:1.7.0.141-2.6.10.1.el6_9"
+            }
+          ],
           "Available": {
             "date": "2024-01-01",
             "kind": "first-observed"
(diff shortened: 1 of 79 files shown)


tests/unit/providers/rhel/test-fixtures/snapshots/rhel:6/cve-2017-3511.json
```diff
@@ -18,6 +18,15 @@
       "Description": "An untrusted library search path flaw was found in the JCE component of OpenJDK. A local attacker could possibly use this flaw to cause a Java application using JCE to load an attacker-controlled library and hence escalate their privileges.",
       "FixedIn": [
         {
+          "AdditionalAdvisoryFixes": [
```
willmurphyscode, line 21: What populate additional if there's only the one?

tests/unit/providers/rhel/test_multi_rhsa.py
```diff
@@ -62,27 +65,117 @@ def test_get_version_base(self, version, expected):
         assert _get_version_base(version) == expected
 
 
-class TestBuildVulnerableRange:
-    def test_distinct_bases_python39(self):
-        # the canonical CVE-2024-8088 shape (already deduped + sorted ascending by base)
-        fixes = [_ar("0:3.9.18-3.el9_4.5"), _ar("0:3.9.19-8.el9")]
-        assert _build_vulnerable_range(fixes) == "< 0:3.9.18-3.el9_4.5 || >= 0:3.9.19, < 0:3.9.19-8.el9"
+class TestSameBaseAdditionalAdvisoryFixes:
```
willmurphyscode, line 68: Make sure this has real data behind it - choose real examples.

tests/unit/providers/rhel/test-fixtures/snapshots/rhel:9/cve-2022-1923.json
```diff
@@ -18,6 +18,15 @@
       "Description": "A flaw was found in GStreamer. An integer overflow can lead to a heap-based buffer overflow in the mkv demuxer when processing a specially crafted Matroska/WebM file using bzip decompression. This vulnerability can result in application crash, memory corruption, and code execution.",
       "FixedIn": [
         {
+          "AdditionalAdvisoryFixes": [
+            {
+              "Advisory": {
+                "ID": "RHSA-2023:2260",
+                "Link": "https://access.redhat.com/errata/RHSA-2023:2260"
+              },
+              "Version": "0:1.18.4-6.el9"
```
willmurphyscode, line 27: Can the vunnel side bucket these into RHEL minor versions ahead of time? Like just capturing the raw field in here and then letting grype bucket by it later maybe it should be a separate field.

willmurphyscode: Something something package qualifiers.

PR #1237, opened 2026-06-30
feat(rhel): RHEL minor aware advisory grouping

willmurphyscode:
Previously when two RHSAs covered the same (CVE ID, RHEL major version, RPM name) tuple, Vunnel would take the RHSA with the higher package version (epoch:version-release "EVR") so that downstream version comparison would occasionally produce a false positive (if someone had applied the RHSA with the lower EVR on the RPM) but could never produce a false negative (because any RPM with EVR greater than or equal to the higher RHSA EVR must have the fix applied.) This approach leads to false positives when the lower EVR RHSA has been applied.

Therefore, when confronted with the situation that multiple RHSAs (or RHBAs or RHEAs) apply to the same (CVE ID, RHEL major version, RPM name) tuple, attempt to guess the support channel (ga, eus, els, etc) and RHEL minor version (e.g. the 2 in RHEL 9.2) from the product IDs in RHEL's CSAF VEX formatted advisories.

Other vendors, such as SUSE, simply namespace every minor version completely, such that SLES 15.2 and SLES 15.3 just have different vulnerability sets. However, Red Hat lets fixes accumulate across minor versions, such that if a fix was applied in RHEL 9.4 a separate 9.5 fix is not expected in the data, so that approach is not available here.

Note that this approach depends on parsing RHEL minor version and support stream information out of the product ID field in CSAF VEX JSON. The spec does not support this parsing, and it will probably break at some point in the future. The product ID handling code does presently pass the current production and beta versions of RHEL CSAF VEX product IDs.

### DIFF
--- a/src/vunnel/providers/rhel/csaf_parser.py
+++ b/src/vunnel/providers/rhel/csaf_parser.py
@@ -3,6 +3,7 @@ import logging
 from packageurl import PackageURL
 
 from vunnel.providers.rhel.csaf_client import CSAFClient
+from vunnel.providers.rhel.product_id import parse_product_id
 from vunnel.utils.csaf_types import CSAFDoc
 from vunnel.workspace import Workspace
 
@@ -130,37 +131,50 @@ class CSAFParser:
         return platform_cpe, module, name, version
 
     # note: this is really taking an ar dict because of a circular import issue. It should be taking an AffectedRelease object.
-    def get_fix_info(self, cve_id: str, ar: dict[str, str | None], normalized_pkg_name: str) -> tuple[str | None, str | None]:
+    def get_fix_info(
+        self,
+        cve_id: str,
+        ar: dict[str, str | None],
+        normalized_pkg_name: str,
+    ) -> tuple[str | None, str | None, str | None, list[str]]:
         """
         Given a CVE ID, an affected release object, and the normalized name of the affected package,
         interrogate CSAF RHSA data to look for an advisory that tells us what version of the package is
         fixed, and the modularity of the package if any.
+
+        Returns a (version, module, product_id, channels) tuple. product_id is the matched CSAF full
+        product id (FPI), which -- unlike the version, product_name, or cpe -- encodes the target RHEL
+        minor stream; the caller uses it to recover the per-stream minor for same-base multi-RHSA
+        handling. ``channels`` is the sorted, deduped set of recognized channel tokens -- including the
+        explicit ``"ga"`` token for generally-available FPIs -- computed across ALL FPIs that ship this
+        exact (version, module) build (e.g. ``["ga"]``, ``["ga", "eus"]``, ``["aus", "eus"]``). It is
+        empty only when NO FPI resolved to a recognized channel (i.e. the channel set is unknown).
 
         The `ar` dict is expected to have the following
         """
         fix_id = ar.get("rhsa_id")
         if not fix_id:
             # This affected release does not reference an advisory; assume it isn't fixed
-            return None, None
+            return None, None, None, []
         doc = self.csaf_client.csaf_doc_for_rhsa(fix_id)
         if not doc:
             # There was no document available for the referenced RHSA. This is unexpected,
             # but logged in the csaf client.
-            return None, None
+            return None, None, None, []
 
         # The CSAF spec allows N vulnerabilities per document. Find the first one whose CVE matches
         # the CVE ID we're looking for fix info for.
         vuln = next((v for v in doc.vulnerabilities if v.cve == cve_id), None)
         if not vuln:
             self.logger.trace(f"{cve_id}: {fix_id} CSAF doc does not claim to fix this CVE")  # type: ignore[attr-defined]
-            return None, None
+            return None, None, None, []
 
         # There are multiple remediations in a CSAF doc, including ones that say things like "no fix available".
         # Choose the "vendor_fix" type remediation that corresponds to the advisory ID mentioned in the affected release.
         remediation = next((r for r in vuln.remediations if r.category == "vendor_fix" and r.url and r.url.endswith(fix_id)), None)
         if not remediation:
             self.logger.trace(f"{cve_id} no remediation obj for {fix_id}")  # type: ignore[attr-defined]
-            return None, None
+            return None, None, None, []
 
         # assume that one of the products listed as rebuilt due to the advisory indicates the first fixed
         # version of the package.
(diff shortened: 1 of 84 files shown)


willmurphyscode:
We will have to emit all RHSAs in all appropriate minors to make this work.

src/vunnel/providers/rhel/parser.py
wagoodman, line 663: I'm not sure if there is an issue here, but there might be an inconsistency. `additional_advisories` is gated on `distinct_base_fixes > 1`... but the new advisories table is gated on `distinct_stream_fixes > 1`.

In the case where there is the same-base multi-minor (e.g. `3.9.18-3.el9_2` and `3.9.18-5.el9_4` with one upstream base) the base-fixes collapses to 1 so `additional_advisories` is empty and `AdvisorySummary` lists only the newest RHSA, while advisories lists both.

src/vunnel/providers/rhel/product_id.py
```diff
@@ -0,0 +1,196 @@
...
+"""
+
+from __future__ import annotations
+
+import re
+from typing import NamedTuple
+
+# Dot-separated tokens (old format) or hyphen suffix (new format) that mark a pinned-minor extended
+# maintenance stream. Matched case-insensitively as whole tokens and normalized to lowercase tokens.
+_EXTENDED_MARKERS = ("EUS", "E4S", "AUS", "TUS")
```
wagoodman, line 56: Should this have `ELS`?

src/vunnel/providers/rhel/parser.py
```diff
@@ -897,6 +954,27 @@ def _parse_cve(self, cve_id, content):  # noqa: C901, PLR0912, PLR0915
...
                         record["VulnerableRange"] = artifact.vulnerable_range
 
+                    # Advisories carries the full per-stream fix table for same-base multi-minor RHSAs:
+                    # every distinct fix build (including the one that became the top-level Version),
+                    # each with the target RHEL minor + the channels delivering it, recovered from the
+                    # CSAF product_ids. Channels lists every recognized channel including "ga" for
+                    # generally-available builds (e.g. ["ga"], ["ga","eus"], ["eus"], ["aus","eus"]);
+                    # empty means the channel set is unknown. Emitted only when there are 2+ distinct
+                    # fix streams; additive (OS schema
+                    # 1.1.1) and ignored by older grype, which continues to use Version. A downstream
```
wagoodman:
```suggestion
                    # 1.1.2) and ignored by older grype, which continues to use Version. A downstream
```

src/vunnel/providers/rhel/product_id.py
willmurphyscode: It should not, I think, because ELS is extra time to support a _major_ version. Rev updates comment.

src/vunnel/providers/rhel/parser.py
willmurphyscode: They are technically different decisions. Basically https://github.com/anchore/vunnel/pull/1214 still stands in this PR. I'm open to feedback there.

src/vunnel/providers/rhel/rhsa_provider.py
```diff
@@ -23,6 +25,14 @@ def __init__(  # noqa: PLR0913
         self.module: str | None = module
         self.package: str | None = package  # the raw "package" field from Hydra JSON API
         self.platform_cpe: str | None = None  # the CPE for the platform, if available
+        # The matched CSAF full product id (FPI) that supplied this fix's version. Unlike platform/cpe
+        # it encodes the target RHEL minor stream, so it is what lets us tell same-base per-stream
+        # fixes apart. Populated during the RHSA lookup; may be None for non-CSAF-sourced versions.
+        self.product_id: str | None = product_id
+        # The sorted, deduped set of extended-support channel tokens (e.g. ["aus", "eus"]) that this
+        # exact build is delivered through. Empty means the build is generally available (shipped via a
+        # normal/GA/non-extended channel). Computed across all FPIs for the build during the RHSA lookup.
```
wagoodman: There is a comment here about "empty means the build is GA", but I think the implementation is the opposite of that: GA resolves to an explicit "ga" token, and empty means unknown.

wagoodman: I think the code is correct though relative to grype (worth double checking)... so just a docstring update

tests/unit/providers/rhel/test_rhel.py
```diff
@@ -634,6 +638,23 @@ def test_parse_affected_releases_0(self, mock_cve, tmpdir):
...
+                                minor=None,
+                                channels=[],
+                            ),
+                            StreamAdvisory(
+                                advisory=Advisory(wont_fix=False, rhsa_id=None, link=None, severity=None),
+                                version="7.2-3.el7",
+                                minor=None,
+                                channels=[],
+                            ),
+                        ),
```
wagoodman, line 657: this is a clarifying question, not request for change: what is the downstream impact of both of these having `minor=None`? I think that means a downstream matcher can't map an installed minor onto either of these. If that's true, when why write this at all? (I think I'm missing something)

src/vunnel/providers/rhel/product_id.py
```diff
@@ -0,0 +1,201 @@
...
+        # rest looks like ".0.GA" or ".Z.MAIN.EUS"; split into dot tokens and resolve tri-state.
+        tokens = [t for t in rest.split(".") if t]
+        channel = _channel_from_tokens(tokens)
+        return ProductIdInfo(minor=minor, channel=channel)
+
+    # ELS: extended *lifecycle* support is a major-only stream (no minor in the prefix), but it IS an
+    # extended-support channel. Recognize it from the trailing "-ELS" marker even when no minor parses.
+    if re.search(r"(?:^|-)ELS(?::|$)", prefix, flags=re.IGNORECASE):
+        return ProductIdInfo(minor=None, channel="els")
+
```
wagoodman, line 180: This ELS check runs after the first two regexs fail... so if there is a ELS FPI that caries a minor then this is never reached (thus we drop the els indicator silently). I didn't see a test case for this.

wagoodman: right, but ELS is major only... though we're going to have a fun time with ELC in a few years... which looks like it pins to minors starting with RHEL 8.10 potentially https://access.redhat.com/support/policy/updates/errata

src/vunnel/providers/rhel/csaf_parser.py
```diff
@@ -130,48 +131,118 @@ def platform_module_name_version_from_fpi(self, doc: CSAFDoc, fpi: str) -> tuple
...
+            doc,
+            candidate_full_product_ids,
+            normalized_pkg_name,
+            ar_plat_cpe,
+            version,
+            module,
+        )
+        return version, module, product_id, channels
+
+    def channels_from_fpis(  # noqa: PLR0913
```
wagoodman, line 209: I was trying to find a test case that produced multiple elements (e.g. `["ga","eus"]` or  `["aus","eus"]` ) but couldn't find direct unit tests or this checked indirectly either.

src/vunnel/providers/rhel/product_id.py
```diff
@@ -0,0 +1,201 @@
...
+        minor = int(old_match.group("minor"))
+        rest = old_match.group("rest") or ""
+        # rest looks like ".0.GA" or ".Z.MAIN.EUS"; split into dot tokens and resolve tri-state.
+        tokens = [t for t in rest.split(".") if t]
+        channel = _channel_from_tokens(tokens)
+        return ProductIdInfo(minor=minor, channel=channel)
+
+    # ELS: extended *lifecycle* support is a major-only stream (no minor in the prefix), but it IS an
+    # extended-support channel. Recognize it from the trailing "-ELS" marker even when no minor parses.
+    if re.search(r"(?:^|-)ELS(?::|$)", prefix, flags=re.IGNORECASE):
```
wagoodman: This is subtly missing some cases...
```suggestion
    if re.search(r"(?:^|-)ELS(?:[.:]|$)", prefix, flags=re.IGNORECASE):
```

RHEL 6 ships via `6Server-ELS.EXTENSION` (the `.EXTENSION` repo segment). There is an example of this for `bind` in `test-fixtures/csaf/advisories/2025/rhsa-2025_23414.json`.

Adding a new test case `("6Server-ELS.EXTENSION:bind-32:9.8.2-0.68.rc1.el6_10.17.x86_64", (None, "els"))` to `tests/unit/providers/rhel/test_product_id.py` will show the issue.

tests/unit/providers/rhel/test_product_id.py
```diff
@@ -0,0 +1,182 @@
+import pytest
+
+from vunnel.providers.rhel.product_id import ProductIdInfo, minor_from_dist_tag, parse_product_id
+
+
+class TestParseProductId:
```
wagoodman, line 6: Can you add a case that has multiple digits in the minor version? something like `.el8_10` (the regex works now, adding a case will protect against future alterations silently regressing)

src/vunnel/providers/rhel/product_id.py
wagoodman: addressed in a2ed69f97fe312c22e1a0f28abee833176fd0764

tests/unit/providers/rhel/test_rhel.py
willmurphyscode: `None` is when vunnel can't figure out which minor version it belongs to. the grype transformer will copy this to every minor, which will have the same degree of FPs as are present today.

src/vunnel/providers/rhel/csaf_parser.py
willmurphyscode: https://github.com/anchore/vunnel/pull/1237/changes/a2ed69f97fe312c22e1a0f28abee833176fd0764#diff-97611388fab8b3f97b11a630c6527c4318a31335e61fda2488bbdc15936a2674R215 is the test you're looking for I believe.

wagoodman (approved):

PR #1264, opened 2026-07-16
fix: guard None reference links in wolfi/chainguard secdb parser

### LINKED ISSUE #1263: wolfi/chainguard providers crash with TypeError when secdb secfixes contain IDs without CVE-/GHSA- prefix (e.g. GO- IDs)
## What happened

Running the `wolfi` or `chainguard` provider against a secdb `security.json` whose `secfixes` lists contain a vulnerability ID that is not `CVE-` or `GHSA-` prefixed (for example a Go vulnerability database ID such as `GO-2026-5932`) crashes the provider:

```
ERROR wolfi: error during update: 'NoneType' object is not iterable
  ...
  File ".../vunnel/providers/wolfi/__init__.py", line 66, in update
  File ".../vunnel/providers/wolfi/parser.py", line 189, in get
  File ".../vunnel/providers/wolfi/parser.py", line 132, in _normalize
  File ".../vunnel/providers/wolfi/parser.py", line 64, in build_reference_links
TypeError: 'NoneType' object is not iterable
```

The `chainguard` provider fails identically because it uses the same `SecDBParser` from `vunnel.providers.wolfi.parser`.

## Why

`vunnel.utils.vulnerability.build_reference_links()` returns `None` for IDs it does not recognize (anything not `CVE-`/`GHSA-` prefixed):

https://github.com/anchore/vunnel/blob/main/src/vunnel/utils/vulnerability.py

The wolfi parser's `build_reference_links` calls `urls.extend(...)` on that return value without a guard:

https://github.com/anchore/vunnel/blob/main/src/vunnel/providers/wolfi/parser.py

The `minimos` provider already guards the identical call (`if reference_links:`) and the minimos secdb publishes `GO-` IDs alongside CVE/GHSA aliases today, importing cleanly through that guarded path into the shipping grype database. The wolfi/chainguard parser is strictly less robust than minimos for the same input shape.

## Real-world occurrence

On 2026-07-16 the Chainguard `chainguard` and `wolfi` secdb feeds briefly included `GO-` IDs in secfixes (rolled back the same day while this crash exists); both providers failed with the traceback above on every run in that window.

## Reproduction

Minimal `security.json`:

```json
{
  "apkurl": "{{urlprefix}}/{{reponame}}/{{arch}}/{{pkg.name}}-{{pkg.ver}}.apk",
  "archs": ["x86_64"],
  "reponame": "os",
  "urlprefix": "https://packages.wolfi.dev",
  "packages": [
    { "pkg": { "name": "coreutils", "secfixes": { "0": ["GO-2026-5932"] } } }
  ]
}
```

Place it as the wolfi provider's input `security.json` (or serve it and point `secdb_url` at it) and run `vunnel run wolfi`: the provider fails with `TypeError: 'NoneType' object is not iterable` instead of importing the record.

## Proposed fix

Guard the `None` the same way the minimos provider does, so unknown ID schemes fall back to just the distro security page link. PR incoming.

philroche:
Fixes the `wolfi` and `chainguard` providers crashing with `TypeError: 'NoneType' object is not iterable` when a secdb `secfixes` list contains a vulnerability ID that is not `CVE-` or `GHSA-` prefixed (for example Go vulnerability database IDs such as `GO-2026-5932`).

`vunnel.utils.vulnerability.build_reference_links()` returns `None` for ID schemes it does not recognize, and the wolfi parser (shared by the `chainguard` provider) called `urls.extend(...)` on that value without a guard. The `minimos` provider already guards the identical call, which is why the minimos secdb's `GO-` IDs import cleanly today while the same input shape crashes wolfi/chainguard. This change applies the same guard, so unknown ID schemes fall back to just the distro security page link.

Tests:
- New unit test for `build_reference_links` covering an unrecognized scheme (`GO-`) and the existing `CVE-` behavior.
- Added a `GO-` ID to the `_normalize` fixture so the end-to-end secdb parse covers a mixed CVE/GHSA/GO secfixes list (this reproduces the crash without the fix).

Fixes #1263

### DIFF
--- a/src/vunnel/providers/wolfi/parser.py
+++ b/src/vunnel/providers/wolfi/parser.py
@@ -67,7 +67,12 @@ class Parser(abc.ABC):
     def build_reference_links(self, vulnerability_id: str) -> list[str]:
         urls = []
         urls.append(f"{self.security_reference_url}/{vulnerability_id}")
-        urls.extend(vulnerability.build_reference_links(vulnerability_id))
+        # build_reference_links returns None for IDs it does not recognize
+        # (anything not CVE-/GHSA- prefixed, e.g. GO- IDs); guard it the same
+        # way the minimos provider does so such IDs import without crashing.
+        links = vulnerability.build_reference_links(vulnerability_id)
+        if links:
+            urls.extend(links)
         return urls
 
     @abc.abstractmethod
--- a/tests/unit/providers/wolfi/test_wolfi.py
+++ b/tests/unit/providers/wolfi/test_wolfi.py
@@ -126,7 +126,7 @@ class TestParser:
                 {
                     "pkg": {
                         "name": "coreutils",
-                        "secfixes": {"0": ["CVE-2016-2781"]},
+                        "secfixes": {"0": ["CVE-2016-2781", "GO-2026-5932"]},
                     },
                 },
                 {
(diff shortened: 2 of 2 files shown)


spiffcs (approved):

spiffcs:
Thanks @philroche for the great report and patch. I've kicked off all the CI jobs and am getting the quality gate to run. When that's all 🟢 we'll get this merged 😄

PR #1272, opened 2026-07-21
fix(ubuntu): remove ubuntu-cve-tracker from input if exists

westonsteimel:
The ubuntu-cve-tracker git repo is no longer used by the ubuntu provider and it takes up a massive amount of space in the cache, so ensure it is removed if it exists.


kzantow (approved):
LGTM

PR #1281, opened 2026-07-27
fix(rhel): drop unused skip_if_exists parameter from _sync_cves

### LINKED ISSUE #17: skip_if_exists not passed to centos parser within RHEL parser
The RHEL parser uses the centos parser in order to download results. The centos parser has a `skip_if_exists` attribute that is leveraged by the centos provider based on the runtime config (wired into `centos.Parser.get()`). The same functionality in the RHEL parser is not wired to the internal centos parser instance even though there is a `skip_if_exists` feature for the RHEL provider as a whole. 

Question: is leaving out `skip_if_exists` on the centos parser calls within the RHEL parser intentional? If so, we need some comments as to why to help folks out in the future.

Eljees:
Closes #17

## Answering the TODO

`_sync_cves()` carried `# TODO: ALEX, should skip_if_exists be hooked up here? (currently unused)`. This proposes the answer "no", and removes the unused parameter instead.

Honoring `skip_if_exists` here would mean skipping a CVE because a file already exists on disk, which silently drops upstream updates for every CVE already fetched once.

The round trips it was meant to save are already avoided in a safe way: `_process_minimal_cve()` compares the minimal CVE returned by the API with the copy on disk and only downloads the full CVE when the two differ, or when either the minimal or full file is missing. That is strictly better — it skips the same downloads without ever missing a change.

## Change

- drop the unused `skip_if_exists` parameter from `_sync_cves()` and its single call site in `get()`
- replace the TODO with a comment explaining why the flag does not apply here

`skip_if_exists` remains honored where it does have meaning: `get()` still receives it from `config.runtime.skip_if_exists` and passes it to `_init_rhsa_data()`.

No behavior change — the parameter was never read.

Happy to flip this the other way and wire the flag up instead if you'd prefer that semantics.

### DIFF
--- a/src/vunnel/providers/rhel/parser.py
+++ b/src/vunnel/providers/rhel/parser.py
@@ -208,8 +208,12 @@ class Parser:
                 with open(os.path.join(dir_path, file), encoding="utf-8") as fp:
                     yield orjson.loads(fp.read())
 
-    # TODO: ALEX, should skip_if_exists be hooked up here? (currently unused)
-    def _sync_cves(self, skip_if_exists=False, do_full_sync=True):  # noqa: PLR0915, PLR0912, C901
+    # note: this intentionally does not take skip_if_exists. Skipping a CVE purely because a file
+    # is already on disk would drop upstream updates; instead _process_minimal_cve() compares the
+    # minimal CVE from the API against the copy on disk and only downloads the full CVE when the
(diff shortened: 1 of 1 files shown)


Eljees:
Ping - open since 27 July. This only drops an unused parameter from _sync_cves, so it should be quick to judge either way.

PR #1309, opened 2026-08-18
fixdate: prefer first-observed over an advisory's published date- #1308

CoreyCook8:
Finder.best() ranked inaccurate candidates (an advisory's own published date, kind="advisory", accurate=False) above first-observed dates. Because a published date routinely predates the fix, it sorted earlier and won in both selection branches, so records whose only candidate is the advisory date (e.g. govulndb GO-* records) reported a fix-availability date months before the fix actually shipped.

Rank inaccurate candidates as a true last resort: accurate candidates and strategy results keep their precedence, but an inaccurate candidate is now used only when there is no first-observed date at all.

Add regression tests covering both branches (accurate and not-yet-accurate first-observed) with a candidate that predates the observed fix.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -308,14 +308,14 @@ wheels = [
 
 [[package]]
 name = "gitpython"
-version = "3.1.58"
+version = "3.1.59"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "gitdb" },
 ]
-sdist = { url = "https://files.pythonhosted.org/packages/26/d6/5f358ff283325580c2003a6d953aea18cfe10ae87b46f5ebc80fa3a386dc/gitpython-3.1.58.tar.gz", hash = "sha256:621416df10ef3fd0e19fabf9172ddeed0fa704d353d04f194eec56a625a95b22", size = 228498, upload-time = "2026-08-04T15:05:49.47Z" }
+sdist = { url = "https://files.pythonhosted.org/packages/ca/dc/126b28e76b24a9268ba931ad3e012f71ebdadf62fd9f17758f7074bb0b20/gitpython-3.1.59.tar.gz", hash = "sha256:0a1475cfdc38a5bfba1a3e9a4a9da52a39749ecec322b772915c019f94e5b7e4", size = 230445, upload-time = "2026-08-10T12:03:20.271Z" }
 wheels = [
-    { url = "https://files.pythonhosted.org/packages/ec/0c/9d8752098bc442f0726e64aa6135940b3a96809915d1aa4206c1bb97881d/gitpython-3.1.58-py3-none-any.whl", hash = "sha256:d331e722577f0fd7fc1f857419b3ecc07af66282b933d2a4d95f84a042fdd50f", size = 220183, upload-time = "2026-08-04T15:05:48.025Z" },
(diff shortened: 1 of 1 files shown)


CoreyCook8:
👋 @wagoodman @willmurphyscode @kzantow @spiffcs (Sorry for the direct ping, not sure if there's a better way to ask for a review) Could someone please take a look at this? I'm noticing that the fix available date was intended to be either the verified date of first available, or the date where the Fix was first present in the database, however we are seeing the published advisory date taking precedence here.

willmurphyscode:
> however we are seeing the published advisory date taking precedence here.

@CoreyCook8 the advisory date is intentionally considered better than the first observed date. For example, consider a distro that has been publishing advisories longer than there have been Grype DB's. In that case, the advisory date is _better_.

We're not trying to compute the first time a fix is available _according to grype_. We're trying to compute the first time a it was _possible in principle_ to upgrade to a patched version, and for distros that have an Advisory date, like the date an RHSA was published, that's a good approximation.

Can you help me understand a little better why you think this strategy is wrong? Do you have particular examples where we're showing the wrong date?

CoreyCook8:
Hey @willmurphyscode there were a series of GO vulnerabilities that had a fixed date with the published date. For example GO-2026-5026. It was fixed ~2 weeks ago, but the fix available date given was 5-22-26

CoreyCook8:
Hey @willmurphyscode any updates on this? Can I provide some more information or help in any way?

willmurphyscode:
Hi @CoreyCook8 thanks for the reply. I'm not sure this is the right change _in general_, probably because the term "advisory" means too many things. In the context of fixed dates, what "advisory" means is like when a distro publishes an advisory, like an RHSA, that is a good proxy for a fixed version becoming available. (Hey, here's all these new RPMs you should download because they fix all these CVEs!). But in the context of GHSA and OSV, the advisory is a continually updated record of disclosure and fixes, so it's probably the wrong choice in many cases to use its date.

But first-observed is usually less accurate than this - it's our fallback of last resort. What it means in practice is, "the first time we published a Grype DB with this fix in it." Since the GO provider was turned up in June, it will give mid June for the vast majority of GO issues, which isn't a net improvement.

Since the set of Go releases is finite and not huge, and the GO fixed dates that are causing trouble are really only attached to stdlib and extended stdlib (golang.org/x/... stuff), probably a better fix is to have vunnel figure out the release date of every go version, and then just look up the fixed version in that map.

In general, it would be easier to change what the go provider passes into the fix date finder. Changes to the fix date finder itself have to be no worse across all the providers, which is a lot of testing to do. Also, the [OSV fixed date utility](https://github.com/anchore/vunnel/blob/main/src/vunnel/utils/osv.py#L9) is in play here - I think the right way to fix this is to pass `accurate=True` extra candidates computed by the Go version release dates in [here](https://github.com/anchore/vunnel/blob/e5681bf2d47c3beacfcc0cd3bf677d9f08053f56/src/vunnel/providers/govulndb/parser.py#L119)

CoreyCook8:
Hey @willmurphyscode based on your feedback here I opened https://github.com/anchore/vunnel/pull/1334.

It's a more go-specific approach which should produce much more accurate dates. Would you be able to take a look at that when you have a moment?

PR #1327, opened 2026-08-31
SLES CSAF provider

### LINKED ISSUE #626: Enhance SLES provider to pull in data on unfixed packages

The SLES provider should be enhanced to pull in the OVAL data stating that a package is affected but not fixed

**Why is this needed**:

This would allow making SLES a comprehensive distro in grype and would eliminate a [large number of false positives](https://github.com/anchore/grype/issues/created_by/sekveaja)

willmurphyscode:
Results in the new labels at https://github.com/anchore/vulnerability-match-labels/pull/208 

Fixes https://github.com/anchore/vunnel/issues/626

See details in src/vunnel/providers/sles/README.md


wagoodman (approved):

willmurphyscode:
This approach has some false positives that aren't found on main. For example, in https://ftp.suse.com/pub/projects/security/csaf-vex/cve-2019-13057.json - libldap-data is listed as `known_affected` but a fixed version is published for its upstream. So this needs a little bit more investigation before we merge it.

PR #1334, opened 2026-09-09
Use reliable GO sources for fix available dates

CoreyCook8:
## Description

Improve Go vulnerability fix-availability dates by deriving high-confidence fix-date candidates from the actual Go release/module version that introduced the fix.

Go vuln DB OSV records do not include per-fix availability dates, and the top-level `published` field represents vulnerability disclosure/publication timing rather than when the fixed version became available. Falling back to first-observed is
also inaccurate for this provider because most Go records date to when the provider was first enabled.

This change keeps the shared fix-date finder behavior unchanged and updates only the `govulndb` provider to pass provider-specific `accurate=True` candidates into `osv.patch_fix_date(...)`.

For fixed versions:

- `stdlib` and `toolchain` dates are resolved from the corresponding `golang/go` release tag.
- Go module dates, including `golang.org/x/...`, are resolved from the Go module proxy version metadata.
- If no release date can be resolved, behavior falls back to the existing advisory/first-observed candidate flow.

## Why

This makes Go fix dates line up with when the fixed Go/toolchain/module version was actually released, instead of using GHSA/OSV publication dates or first-observed dates.

This is especially important for historical Go vulnerabilities where first-observed commonly reflects provider turn-up time rather than real fix availability.

## Testing

Added focused tests for:

- Go stdlib/toolchain version-to-tag normalization.
- Go module proxy release-date lookup.
- stdlib release tag date lookup.
- release-date lookup caching.
- `accurate=True` extra candidate generation.
- provider-level behavior showing Go release dates are preferred over first-observed fallback.

Replaces the fix offered in https://github.com/anchore/vunnel/pull/1309 based on the feedback on that PR.

### DIFF
--- a/src/vunnel/providers/govulndb/go_release_dates.py
+++ b/src/vunnel/providers/govulndb/go_release_dates.py
@@ -42,6 +42,11 @@ GO_GIT_URL = "https://go.googlesource.com/go"
 # module names in the go vuln db that map to Go toolchain release tags rather
 # than proxy module versions
 STDLIB_MODULES = {"stdlib", "toolchain"}
+USER_AGENT = "vunnel/1.0 (govulndb-provider; +https://github.com/anchore/vunnel)"
+
+
+class GoReleaseDateLookupError(RuntimeError):
+    """Raised when an authoritative Go release-date source cannot be queried."""
 
 
 class GoReleaseDateOverlay:
@@ -52,9 +57,18 @@ class GoReleaseDateOverlay:
     tuple never re-hit the network.
     """
 
-    def __init__(self, logger: logging.Logger | None = None):
+    def __init__(
+        self,
+        logger: logging.Logger | None = None,
+        timeout: int = http.DEFAULT_TIMEOUT,
+        retries: int = 3,
+        backoff_in_seconds: int = 3,
+    ):
         self._cache: dict[tuple[str, str], date | None] = {}
         self.logger = logger if logger is not None else logging.getLogger(self.__class__.__name__)
+        self.timeout = timeout
+        self.retries = retries
+        self.backoff_in_seconds = backoff_in_seconds
 
     def lookup(self, module: str, version: str) -> date | None:
         """Return the release date for a Go (module, version), or None on miss."""
@@ -96,8 +110,7 @@ class GoReleaseDateOverlay:
         try:
             return orjson.loads(resp.content)
         except orjson.JSONDecodeError:
-            self.logger.warning(f"failed to parse JSON from {url}")
-            return None
+            raise GoReleaseDateLookupError(f"failed to parse JSON from {url}") from None
 
     def _get_gitiles_json(self, url: str) -> dict[str, Any] | None:
         # gitiles prefixes JSON with a `)]}'` XSS-guard line that must be stripped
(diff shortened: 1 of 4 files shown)


src/vunnel/providers/govulndb/go_release_dates.py
```diff
@@ -0,0 +1,388 @@
...
+
+    def _fetch_json(self, url: str, strip_xss_guard: bool) -> dict[str, Any] | None:
+        """GET `url` and decode its JSON body, or return None and log why.
+
+        Deliberately total: every failure mode an upstream can present -- refused
+        connection, timeout, 5xx, an HTML error page served as 200 -- becomes a
+        missing date rather than an exception, because the caller has a usable
+        fallback and the run does not.
+        """
+        try:
```
wagoodman, line 170: with `retries=2, timeout=10, backoff=3` is 3 attempts plus 3s and 6s backoff, so ~39s per blackholed URL and ~9s per refused one. Times 101 URLs that's 15 to 66 minutes of pure sleeping added to a provider that otherwise just reads a zip. 

The module docstring at promises this can't happen: 

> This overlay never raises: a lookup that cannot be answered -- host down,
garbage body, unrecognized version -- returns None and the caller falls back to
the advisory's published date. Fix dates are an enrichment, and losing one must
not cost the whole provider run.

Sustained 429s might be worse, `MAX_RATE_LIMIT_WAIT` is 300s per wait.

src/vunnel/providers/govulndb/go_release_dates.py
```diff
@@ -0,0 +1,388 @@
...
+    whichever source answered, and matches how every other date in vunnel reads.
+    """
+    if value.tzinfo is None:
+        return value.date()
+    return value.astimezone(UTC).date()
+
+
+def go_extra_candidates(
+    overlay: GoReleaseDateOverlay,
+    aliases: Iterable[str] | None = None,
```
wagoodman, line 365: a default of `aliases=None`... which `should_resolve` maps to "resolve everything". So a caller that forgets the second arg silently gets the 3375-request version.... `usn_extra_candidates` fails closed here, this fails open.

tests/unit/providers/govulndb/test_go_release_dates.py
```diff
@@ -0,0 +1,432 @@
...
+@pytest.fixture
+def overlay():
+    return GoReleaseDateOverlay(logger=logging.getLogger("test"))
+
+
+def stub_http(monkeypatch, handler):
+    """Replace the http layer the overlay calls, recording the urls requested."""
+    calls = []
+
+    def fake_get(url, logger, **kwargs):
```
wagoodman, line 36: this `fake_get` drops the kwargs on the floor... that means the `status_handler` never runs in any test, so the "404/410 aren't worth retrying" decision is untested, and `timeout=10` is unasserted so the "deliberately not download_timeout" contract in `parser.py:52-54` isn't enforced either:

>  note: deliberately not download_timeout -- that is sized for the vulndb zip, and these are single-document metadata reads. With skip_download the overlay serves the committed release table and nothing else, so the run stays offline.

A regression that made 404s retry with backoff against `proxy.golang.org` would pass the full suite.

src/vunnel/providers/govulndb/parser.py
wagoodman, line 43: we should now be including the proxy URL as well here

wagoodman (commented):
There is one thing I think is worth highlighting, and might be a change request: there is a new network dependency inside a per-record loop.

The overlay does per-`(module, version)` HTTP from inside a `patch_fix_date` candidate callback. Nothing in this repo did that before... something like the existing `usn_fixdate_overlay.py` is entirely offline.

`_cached` bounds each *URL* to one lookup, but there's no cross-URL state, so a host outage gets re-learned.

wagoodman (approved):

PR #1350, opened 2026-09-18
fix(http): retry downloads when the connection drops mid-body

willmurphyscode:
`get(..., stream=True)` returns once headers arrive, so its retry loop covered only connect, TLS and status. The caller drained the body outside the loop, so a connection dropped mid-download got no retries at all.

`download_to_file` drains inside the loop, delegating the request to `get` so pooling, rate limiting and backoff are unchanged. Bytes stage in a sibling `.part` and are renamed on success, so a failure leaves nothing truncated. Every streaming download outside the sles provider now uses it.

No resume or ETag revalidation; failed large file downloads are still retried entirely. They are just now retried at all.


tests/quality/vulnerability-match-labels
wagoodman, line 1: was this intentional?

wagoodman (approved):

